Issue #1, opened 2021-01-29
isElementNode does not enforce `children` property


e1himself:
Hi! 

I have noticed that `isElementNode` does not enforce the `children` property to be defined on an object:

https://github.com/prezly/slate/blob/31a471e923a3c99ac5d2aa57535c42bc8949f0fd/packages/slate-types/src/nodes/ElementNode.ts#L10-L12

Is this intentional?

Thanks!

kamilmielnik:
> Is this intentional?

Checking `children` was indeed unintentionally missing. Thank you for the report!

Fixed by https://github.com/prezly/slate/pull/2

PR #2, opened 2021-02-02
isElementNode does not enforce `children` property

kamilmielnik:
Fixes https://github.com/prezly/slate/issues/1

### DIFF
--- a/packages/slate-types/src/nodes/ElementNode.ts
+++ b/packages/slate-types/src/nodes/ElementNode.ts
@@ -1,6 +1,6 @@
 import { isObject } from '../lib';
 
-import TextNode from './TextNode';
+import TextNode, { isTextNode } from './TextNode';
 
 export default interface ElementNode<T extends string = string> extends Record<string, unknown> {
     children: (ElementNode<string> | TextNode)[];
@@ -8,5 +8,12 @@ export default interface ElementNode<T extends string = string> extends Record<s
 }
 
 export const isElementNode = (value: any): value is ElementNode => {
-    return isObject(value) && typeof value.type === 'string' && value.type.length > 0;
+    return (
+        isObject(value) &&
+        typeof value.type === 'string' &&
+        value.type.length > 0 &&
+        Array.isArray(value.children) &&
+        value.children.length > 0 &&
+        value.children.every((child) => isTextNode(child) || isElementNode(child))
+    );
 };


packages/slate-types/src/nodes/ElementNode.ts
```diff
@@ -1,12 +1,19 @@
...
 }
 
 export const isElementNode = (value: any): value is ElementNode => {
-    return isObject(value) && typeof value.type === 'string' && value.type.length > 0;
+    return (
+        isObject(value) &&
+        typeof value.type === 'string' &&
+        value.type.length > 0 &&
+        Array.isArray(value.children) &&
+        value.children.length > 0 &&
```
e1himself, line 16: :thinking: Does the spec actually require the list to have 1+ items?

kamilmielnik: The Slate spec - [no](https://github.com/ianstormtaylor/slate/blob/0513539/packages/slate/src/interfaces/node.ts#L386).

But I like to think that "our spec" does. 

I mean: it's not really possible to create a node with empty `children` array because Slate will automatically normalize it by inserting `{ text: "" }` to that array. We've had various normalization bugs, but I've never seen an empty `children` array.

Do you think this check does not belong here?

e1himself: So I'd say the official spec should be respected over de-factor invariants that might be just temporary implementation detail coincidence.

I'm afraid that inventing this requirement here will make any dependent implementations lock themselves into expecting this expectation. While this can cause problems in future, when the official Slate editor implementation changes to not include empty text nodes (still being 100% aligned with spec).

kamilmielnik: > This can cause problems in future, when the official Slate editor implementation changes to not include empty text nodes (still being 100% aligned with spec).

But what do you consider to be the spec? Docs? Code? They're both likely to change.

Anyway, according to point 1 in here: https://docs.slatejs.org/concepts/10-normalizing#built-in-constraints:

> All Element nodes must contain at least one Text descendant.

We can be very strict about checking these things but we don't have to. 
I'd say the current check is good enough.

e1himself: Alright, thanks!

PR #18, opened 2021-04-05
[DEV-2184] [@prezly/slate-renderer] Prevent the same embed scripts being loaded multiple times

oleg-semyonov:
Attempted fix for this issue:
https://linear.app/prezly/issue/DEV-2184/errors-in-console-when-browsing-a-page-with-multiple-videos

### DIFF
--- a/packages/slate-renderer/src/elements/Embed/lib/injectOembedMarkup.ts
+++ b/packages/slate-renderer/src/elements/Embed/lib/injectOembedMarkup.ts
@@ -13,12 +13,20 @@ const injectOembedMarkup = ({ oembed, onError, target }: Parameters): void => {
     container.innerHTML = oembed.html || '';
     const embedScripts = Array.from(container.getElementsByTagName('script'));
 
-    embedScripts.forEach((embedScript) => {
-        const script = document.createElement('script');
-        copyScriptAttributes(embedScript, script);
-        script.addEventListener('error', onError);
+    const currentScriptSources = Array.from(document.getElementsByTagName('script')).map((script) => (
+        script.getAttribute('src')
+    ));
 
-        document.body.appendChild(script);
+    embedScripts.forEach((embedScript) => {
+        // Prevent same scripts from loading multiple times
+        if (!currentScriptSources.includes(embedScript.getAttribute('src'))) {
+            const script = document.createElement('script');
+            copyScriptAttributes(embedScript, script);
+            script.addEventListener('error', onError);
+    
+            document.body.appendChild(script);
+        }
+
         // Remove the original script so it's not loaded twice.
         embedScript.remove();
     });
--- a/packages/slate-renderer/src/elements/Embed/lib/injectOembedMarkup.ts
+++ b/packages/slate-renderer/src/elements/Embed/lib/injectOembedMarkup.ts
@@ -13,9 +13,9 @@ const injectOembedMarkup = ({ oembed, onError, target }: Parameters): void => {
     container.innerHTML = oembed.html || '';
     const embedScripts = Array.from(container.getElementsByTagName('script'));
 
-    const currentScriptSources = Array.from(document.getElementsByTagName('script')).map((script) => (
-        script.getAttribute('src')
-    ));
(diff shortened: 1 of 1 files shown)


packages/slate-renderer/src/elements/Embed/lib/injectOembedMarkup.ts
```diff
@@ -13,12 +13,20 @@ const injectOembedMarkup = ({ oembed, onError, target }: Parameters): void => {
     container.innerHTML = oembed.html || '';
     const embedScripts = Array.from(container.getElementsByTagName('script'));
 
+    const currentScriptSources = Array.from(document.getElementsByTagName('script')).map((script) =>
```
kamilmielnik, line 16: _this comment is in an unrelated place_

https://linear.app/prezly/issue/DEV-2184#comment-826b8e56

>  Though I couldn't find a way to test that it indeed works. I believe I will have to wait for @kamilmielnik to return and help with that

Yes, there is no easy way to test it. I've created an issue for it: https://github.com/prezly/slate/issues/21

What you can do now is:
1. Set up the `slate` repository locally
2. Make sure to build it
3. Use local path (`file:`) in [`package.json` of your project](https://github.com/prezly/website-nextjs/blob/main/package.json#L14) to specify path to `@prezly/slate-renderer` package you've built locally - see [npm docs](https://docs.npmjs.com/cli/v7/configuring-npm/package-json#local-paths)

kamilmielnik (approved):
Code looks good but I don't know how to reproduce the issue, therefore I don't know what is the source of the issue, therefore I could not test this PR.

Please make sure to test it before merging.

In case of any issues - please, let me know :)

packages/slate-renderer/src/elements/Embed/lib/injectOembedMarkup.ts
oleg-semyonov: Thanks for a tip, will try to test it like that!

packages/slate-renderer/src/elements/Embed/lib/injectOembedMarkup.ts
```diff
@@ -13,12 +13,20 @@ const injectOembedMarkup = ({ oembed, onError, target }: Parameters): void => {
...
+        script.getAttribute('src'),
+    );
+
     embedScripts.forEach((embedScript) => {
-        const script = document.createElement('script');
-        copyScriptAttributes(embedScript, script);
-        script.addEventListener('error', onError);
+        // Prevent same scripts from loading multiple times
+        if (!currentScriptSources.includes(embedScript.getAttribute('src'))) {
```
kamilmielnik, line 22: :x: 

So I have to admit to a terrible overlook from my side... I realized it only after reading [your comment in linear](https://linear.app/prezly/issue/DEV-2184#comment-44026fc9).

Please create an html file with the following content and open it in your browser. Then open up developer tools, refresh the page and answer this question: how many times has `https://cdn.jsdelivr.net/lodash/4/lodash.min.js` been loaded?

```html
<!DOCTYPE html>
<html>
<head>
  <script src="https://cdn.jsdelivr.net/lodash/4/lodash.min.js"></script>
  <script src="https://cdn.jsdelivr.net/lodash/4/lodash.min.js"></script>
</head>
<body>
  <script src="https://cdn.jsdelivr.net/lodash/4/lodash.min.js"></script>
  <script src="https://cdn.jsdelivr.net/lodash/4/lodash.min.js"></script>
</body>
</html>
```

kamilmielnik (changes requested):

packages/slate-renderer/src/elements/Embed/lib/injectOembedMarkup.ts
oleg-semyonov: Yeah, it's loaded only one time 🤔 
So the I guess this PR can be safely discarded :)

oleg-semyonov: @kamilmielnik strangely I don't have option to close the pull request, can you do it please?

PR #20, opened 2021-04-09
[DEV-2469] Remove uploadcare-image package

yuriyyakym:
No need to re-publish everything. UplaodcareImage component is now moved to https://github.com/prezly/uploadcare-image


oleg-semyonov (approved):

kamilmielnik:
> No need to re-publish everything. 

Yes, there is.

These lines in `package.json` need to be updated:
- https://github.com/prezly/uploadcare-image/blob/master/package.json#L14
- https://github.com/prezly/uploadcare-image/blob/master/package.json#L17
- https://github.com/prezly/uploadcare-image/blob/master/package.json#L19
- + same things in `package-lock.json`

And then a release should follow (can be a _patch_ version bump). 

Also, please remember to keep GitHub releases in sync with npm releases: https://github.com/prezly/uploadcare-image/releases
No need to copy the old releases over, we can start from the one you'll be about to make. 

cc: @oleg-semyonov, @yuriyyakym

PR #22, opened 2021-04-28
[DEV-2661] Apply the same positioning to both GIFs and regular images

oleg-semyonov:


kamilmielnik (commented):
Linear issue does not include any information on where this problem occurs or how to reproduce it. 

Can you please add it (please use our bug template), so I can verify the fix?

oleg-semyonov:
@kamilmielnik after further research I found that the problem was in the Tailwind CSS reset, which adds `display: block` to `video` tag, making it ignore `text-align` rule from the parent `figure` tag.

I documented it in the Linear card, with examples. Please let me know if this fix should be applied, or if I should close this PR and fix it on my theme (but then this fix might be needed for every new theme)

kamilmielnik:
> or if I should close this PR and fix it on my theme (but then this fix might be needed for every new theme)

By default I'd say that since the problem is not in this repository, there is nothing to fix here. And there is nothing in this repository that would prevent the fix, so no reason to change anything here.

But! This PR actually fixes something (probably the very thing that Tailwind's CSS reset addresses). There is a redundant whitespace below the `<video>` element, which this PR deals with. So let's merge it!

image

packages/slate-renderer/src/components/Media/Media.scss
```diff
@@ -3,19 +3,18 @@
...
 .prezly-slate {
     &-media {
-        &--image {
+        &--image,
+        &--video {
             display: block;
             max-width: 100%;
             height: auto;
             margin: 0 auto;
+        }
```
kamilmielnik, line 12: Since it applies to all `.prezly-slate-media`, we can skip the modifiers

```suggestion
        display: block;
        max-width: 100%;
        height: auto;
        margin: 0 auto;
```

kamilmielnik (approved):

kamilmielnik:
@oleg-semyonov 

> So let's merge it!

image

But not without approval :smile:

packages/slate-renderer/src/components/Media/Media.scss
oleg-semyonov: Sorry, merged the PR before your comment. Will apply this fix separately.

PR #25, opened 2021-07-07
Update .prettierignore config to not verify builds

yuriyyakym:
`dist` directories are not pushed to the repository. Nevertheless build fails because some of build fails are not prettified.


yuriyyakym:
@kamilmielnik just FYI

PR #32, opened 2021-10-19
[MT-4401] Chore - Upgrade Slate to v0.66

e1himself:


lerna.json
```diff
@@ -5,5 +5,5 @@
     "publishConfig": {
         "access": "public"
     },
-    "version": "0.1.8"
+    "version": "0.2.0-alpha.0"
```
kamilmielnik, line 8: I think this version bump does not belong in this PR

kamilmielnik: Right!

packages/slate-editor/src/components/Portals/BasePortalV2.tsx
```diff
@@ -75,7 +75,22 @@ const BasePortalV2: FunctionComponent<Props> = ({
         if (rect === null) {
             setReferenceElement(null);
         } else {
-            const clientRect = { x: 0, y: 0, toJSON: () => null, ...rect };
+            /**
+             * We have to manually re-create a ClientRect-shape object instead of `...rect`,
+             * as `DOMRect` object properties are not enumerable.
+             * @see https://github.com/microsoft/TypeScript/issues/9726
```
kamilmielnik, line 81: :open_mouth: 

Good job nailing it down :+1: 

Since this is a open source repo, maybe we can share this as an example of where it is not an [_edge case_](https://github.com/microsoft/TypeScript/issues/9726#issuecomment-240785269)?

kamilmielnik (approved):
:100:

PR #48, opened 2021-11-18
[MT-4485] Refactor - Unbundled build

e1himself:
Rewrite `slate-editor` build from Webpack all-in-one bundle to unbundled set of ES JS (`.mjs`), TS typedefs and CSS files 

- Move `playground` to the root dir (as it is not in fact a sub-package, but more like a dev workspace)
- Introduce `eslint`
- Build TS typedefs with `tsc --build` (faster)
  - Enable `isolatedModules` and force `import type` (to make sure `tsc` and `babel` produce the same output).
    Otherwise `babel` keeps empty modules and type imports in the build, as it does not analyze TS code and does now know which symbol is a type and which is a value. `import type` was introduced specifically for this problem.
- Build JS (`.mjs`) with `babel`
- Build SVG with `babel`
- Build SCSS with `sass` (via gulp)
- Drop `webpack` and co.

This will simplify build system in the main app repo and speed up cross-repo development.


pestaa:
Build configs can be so much fun amirite

Well done seeing this through @e1himself!

PR #60, opened 2021-12-14
Add Void extension

aspirisen:
Ticket: https://linear.app/prezly/issue/MT-4521/selecting-divider-node-with-mouse-and-then-pressing-enter-does-not

Demo:

There is an issue with image component, on Down it skips the image, but on up it focuses on the image label text

### DIFF
--- a/packages/slate-editor/src/modules/editor-v4-void/VoidExtension.tsx
+++ b/packages/slate-editor/src/modules/editor-v4-void/VoidExtension.tsx
@@ -9,21 +9,17 @@ export const VoidExtension = (): Extension => ({
     onKeyDown: (event, editor) => {
         const [currentNode] = EditorCommands.getCurrentNodeEntry(editor) ?? [];
 
-        if (!currentNode || !editor.selection) {
-            return;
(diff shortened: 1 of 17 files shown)


packages/slate-editor/src/modules/editor-v4-void/VoidExtension.tsx
```diff
@@ -0,0 +1,26 @@
...
+export const VoidExtension = (): Extension => ({
+    id: VOID_EXTENSION_ID,
+    onKeyDown: (event, editor) => {
+        const [currentNode] = EditorCommands.getCurrentNodeEntry(editor) ?? [];
+
+        if (!currentNode || !editor.selection || !Editor.isVoid(editor, currentNode)) {
+            return;
+        }
+
+        event.preventDefault();
```
e1himself, line 16: Don't we want to prevent default only if actually handle it?

e1himself (approved):
Amazing work @aspirisen! :raised_hands:

e1himself (approved):
:+1:

PR #62, opened 2021-12-16
Fix - Selecting all caption in Image block and pressing Backspace removes whole block

aspirisen:
Ticket: https://linear.app/prezly/issue/MT-4531/selecting-all-caption-in-image-block-and-pressing-backspace-removes

Demo:

### DIFF
--- a/packages/slate-editor/src/modules/editor-v4-image/ImageExtension.tsx
+++ b/packages/slate-editor/src/modules/editor-v4-image/ImageExtension.tsx
@@ -15,7 +15,6 @@ import {
     createImageCandidate,
     getAncestorAnchor,
     isDeleting,
-    isDeletingBackward,
     normalizeChildren,
     normalizeImageCandidate,
     normalizeRedundantImageAttributes,
@@ -99,15 +98,6 @@ const ImageExtension = ({
                 event.preventDefault();
                 event.stopPropagation();
                 return;
-            }
-
-            if (isDeletingBackward(event) && EditorCommands.isSelectionAtBlockStart(editor)) {
-                EditorCommands.removeNode(editor, {
-                    at: nodeEntry[1],
-                    match: isImageNode,
-                });
-                event.preventDefault();
-                event.stopPropagation();
             }
         }
     },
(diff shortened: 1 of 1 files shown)


packages/slate-editor/src/modules/editor-v4-image/ImageExtension.tsx
```diff
@@ -100,15 +99,6 @@ const ImageExtension = ({
...
             }
-
-            if (isDeletingBackward(event) && EditorCommands.isSelectionAtBlockStart(editor)) {
-                EditorCommands.removeNode(editor, {
-                    at: nodeEntry[1],
-                    match: isImageNode,
-                });
-                event.preventDefault();
-                event.stopPropagation();
-            }
```
e1himself, line 111: Does this mean it's not possible to delete the block by pressing backspace?

Scenario:

1. I have an image block
2. I have a text with paragraph after it
3. I put my cursor to the end of the paragraph
4. I press Backspace multiple types
5. I expect the paragraph contents to be deleted one character after another
6. Once the paragraph is gone I expect my cursor to:
   - Either remove the image block and make the document empty
   - Or move to the image caption text and start deleting it. 
     Once the caption is empty, my next Backspace pressing should remove the image block

How this scenario will work after this change?

aspirisen: @e1himself when the caption is empty and you press Backspace the image block will be removed and the selection will go to the block bellow

e1himself: Good! 

What about step 6 from the scenario above? Will it start deleting the caption or will it remove the image right away if I continue pressing Backspace after deleting the paragraph following the image block?

aspirisen: hm, i will take a look

e1himself: @aspirisen so what will be the behavior now?

aspirisen: @e1himself

e1himself: For future reference: we'll follow up on improving the behavior with backspace and void elements in MT-4532 and MT-4533.

e1himself (approved):
Good job!

PR #71, opened 2022-01-12
[MT-4437] Feature - Text alignment

e1himself:
Peek 2022-01-12 20-14


packages/slate-editor/src/modules/editor-v4-rich-formatting/components/Lists/ListItemTextElement.tsx
```diff
@@ -18,13 +18,13 @@ export const ListItemTextElement: FunctionComponent<Props> = ({
     element,
     ...props
 }) => (
-    <div
+    <span
```
e1himself, line 21: Made this an inline element to have it play well with `list-style-position: inside` with non-default alignment modes.

packages/slate-editor/src/modules/editor-v4-paragraphs/lib/normalizeRedundantParagraphAttributes.ts
```diff
@@ -1,10 +1,15 @@
 import { EditorCommands } from '@prezly/slate-commons';
+import type { ParagraphNode } from '@prezly/slate-types';
 import { isParagraphNode } from '@prezly/slate-types';
 import type { Editor, NodeEntry } from 'slate';
 
-import { createParagraph } from './createParagraph';
+const shape: { [P in keyof ParagraphNode]: true } = {
```
aspirisen, line 6: nit. you can use Record here

packages/slate-editor/src/modules/editor-v4-rich-formatting/components/RichFormattingMenu.tsx
```diff
@@ -30,11 +31,18 @@ import { ElementType } from '../types';
...
 
 interface Props {
+    alignmentControls: boolean;
     containerRef: RefObject<HTMLElement>;
+    defaultAlignment: Alignment;
     parameters: RichFormattingExtensionParameters;
 }
 
-export const RichFormattingMenu: FunctionComponent<Props> = ({ containerRef, parameters }) => {
+export const RichFormattingMenu: FunctionComponent<Props> = ({
```
aspirisen, line 40: nit. FC or FunctionalComponent have some disadvantages https://github.com/facebook/create-react-app/pull/8177

e1himself: Interesting.

So do you think we should stop using `FunctionalComponent` for defining components?

packages/slate-editor/src/modules/editor-v4-paragraphs/lib/normalizeRedundantParagraphAttributes.ts
e1himself: Thanks! 

Done in 48c535c09c114d0947cbe7ccbb6081e823989a09

packages/slate-editor/src/modules/editor-v4-rich-formatting/components/RichFormattingMenu.tsx
aspirisen: in my opinion yes, we could use `function Component(props: Props)` or `function Component(props: React.PropsWithChildren<Props>)`

e1himself: I actually love it! Will try to adopt this format for a few new components before deciding to go all-in. Thanks!

PR #86, opened 2022-01-26
[MT-4590] Feature - [+] button in heading elements

e1himself:
- Make the [+] button appear in empty headings too


aspirisen:
@e1himself could you also take a look on this PR

packages/slate-editor/src/modules/editor-v4-floating-add-menu/lib/shouldShowMenuButton.ts
```diff
@@ -12,7 +12,11 @@ export function shouldShowMenuButton(editor: Editor): boolean {
 
     const [currentNode] = EditorCommands.getCurrentNodeEntry(editor) || [];
 
-    if (!isParagraphNode(currentNode) || EditorCommands.hasVoidElements(editor, currentNode)) {
+    if (
+        !isParagraphNode(currentNode) ||
+        isHeadingNode(currentNode) ||
```
e1himself, line 17: I don't understand this condition. Isn't `isHeadingNode` effectively a noop? Because `!isParagraphNode` already includes a case when the node is a heading node.

packages/slate-editor/src/modules/editor-v4-floating-add-menu/FloatingAddMenu.module.scss
```diff
@@ -0,0 +1,61 @@
+@use "sass:math";
+@import "styles/variables";
+
+.editor-v4-floating-add-menu {
```
e1himself, line 4: Why not simplify this to just `.menu`? :)

packages/slate-editor/src/modules/editor-v4-floating-add-menu/lib/shouldShowMenuButton.ts
e1himself: I think it should be changed to `!(isParagraphNode(currentNode) || isHeadingNode(currentNode) || hasVoidNodes(...))`.

Or even better split it into two conditions to make things clear:

```ts
if (!isParagraphNode(currentNode) && !isHeadingNode(currentNode)) {
   return false;
}
if (hasVoidNodes(...)) {
   return false.
}
```

aspirisen:
@e1himself could you take a look again

e1himself (commented):
Code-wise and functionality-wise, everything's good. Good job Dmitry :+1: 

However, I think the UI/UX part can be improved. I'd recommend you checking with Hugo on it.

Specifically, I think the placeholder text color in h1 and h2 variations is too heavy. I feel like the color should be lighter.

aspirisen:
@e1himself there was comment from Lilian to make placeholders bold https://linear.app/prezly/issue/MT-4590#comment-37e3b7d7

e1himself:
@aspirisen I see. But I still think it's worth checking with Hugo. Feel free to ignore this comment and just merge, if you are absolutely sure it's good.

PR #87, opened 2022-01-28
[MT-4450] Creating a Link does not focus URL input

aspirisen:
Ticket: https://linear.app/prezly/issue/MT-4450/editor-creating-a-link-does-not-focus-url-input

with scroll:

### DIFF
--- a/packages/slate-editor/src/components/Portals/BasePortalV2.scss
+++ b/packages/slate-editor/src/components/Portals/BasePortalV2.scss
@@ -9,7 +9,7 @@
     z-index: $highest-known-z-index + 1;
 
     &--uninitialized {
-        visibility: hidden;
+        opacity: 0;
     }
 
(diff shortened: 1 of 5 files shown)


e1himself (changes requested):
@aspirisen 

I've tested this and found a problem, when there's a lot of content above the paragraph I'm trying to link text in. The scroll jumps up, leaving the link toolbar below the viewport. I'm sure this is a regression, as this does not happen when I do the same test on master. 

See this attached gif:

Peek 2022-02-01 16-22

aspirisen:
@e1himself nice catch 👍

e1himself (approved):
Works well now! Good job :+1: 

I've also pushed a change to make Link menu positioning match Rich Formatting toolbar: 7e6f272e097888d601da71fe124c209a5c8c897e (that's my overlook from #70)

PR #88, opened 2022-01-28
[MT-4535] Show add button for empty block including spaces

aspirisen:
Ticket: https://linear.app/prezly/issue/MT-4535/pressing-enter-before-a-whitespace-creates-a-seemingly-empty-line


packages/slate-commons/src/commands/isNodeEmpty.ts
```diff
@@ -12,5 +17,13 @@ export function isNodeEmpty(editor: Editor, node: Node): boolean {
         );
     }
 
+    if (options?.shouldTrim) {
+        const str = Node.string(node);
+
+        if (str.length && str.trim() === '') {
+            return true;
+        }
+    }
```
e1himself, line 26: Hm... I think there might be a problem here. `Editor.isEmpty` is smarter and does additional checks to cover void nodes (void nodes always generate an empty string for `Node.string()`, but cannot be considered empty).
 
The code here will return `true` for a void node, which I believe, can lead bugs. Right?

I suggest, at least, adding a check to return `false` for void elements:

```tsx
if (editor.isVoid(node)) {
    return false; 
}
```

Also `Node.string()` will also return an empty string for an image element with empty caption, which might lead to bugs too. So it's better to cover that.

Am I overthinking it?

aspirisen: You are right, the logic of this function could be confusing for future usages. 
What about moving this logic into this file https://github.com/prezly/slate/blob/master/packages/slate-commons/src/commands/isEmptyParagraphElement.ts it works with paragraphs and asumes that the node is a text node?

e1himself: > What about moving this logic

Let's try.

aspirisen:
@e1himself added the ticket https://linear.app/prezly/issue/MT-4670/showing-[]-button-inside-spaces-causes-weird-behaviour

e1himself (approved):
Sorry for jumping in with all the suggestions, but I really cherish the code to be simple-to-read, unless there's a reason to complicate it.

This should be good to merge now :+1:

PR #91, opened 2022-02-01
[MT-4535] Suggestion on minor changes here and there

e1himself:
Suggestion for #88


packages/slate-commons/src/commands/isEmptyParagraphElement.ts
```diff
@@ -1,28 +1,18 @@
 import { isParagraphNode } from '@prezly/slate-types';
 import { Node } from 'slate';
-import type { Editor } from 'slate';
 
-import { isNodeEmpty } from './isNodeEmpty';
-
-export interface IsEmptyParagraphElementOptions {
-    shouldTrim?: boolean;
+interface Options {
+    ignoreWhitespace?: boolean;
```
aspirisen, line 5: imho i am not sure about "ignoreWhitespace" name
i.e. in this example "Hello world" has whitespace in the middle, but it will not be ignored

e1himself: Hm... Good point. I didn't think of this.

Let's call it just `trim` then?

packages/slate-commons/src/commands/isCursorInEmptyParagraph.ts
```diff
@@ -2,17 +2,13 @@ import type { Editor } from 'slate';
 import { Range } from 'slate';
 
 import { getCurrentNodeEntry } from './getCurrentNodeEntry';
-import type { IsEmptyParagraphElementOptions } from './isEmptyParagraphElement';
 import { isEmptyParagraphElement } from './isEmptyParagraphElement';
 
-export interface IsCursorInEmptyParagraphOptions {
-    shouldTrim?: IsEmptyParagraphElementOptions['shouldTrim'];
+interface Options {
```
aspirisen, line 7: I think it is worth to export types that are used in exported functions because they are tight together. 
There can be difficulties when the consumer does not have access to the types that are used in exported entities

e1himself: >  There can be difficulties when the consumer does not have access to the types that are used in exported entities

Let's export them *when* we face such difficulties :)

packages/slate-commons/src/commands/isEmptyParagraphElement.ts
```diff
@@ -1,28 +1,39 @@
...
+    options?: Options,
 ): boolean {
     if (!isParagraphNode(node) || !node) {
         return false;
     }
 
-    let nodeEmpty = isNodeEmpty(editor, node);
+    if (hasVoidElements(editor, node)) {
+        return false;
+    }
```
e1himself, line 19: @aspirisen I've also realized that we should take into account void nodes (as there can be inline voids too): 58abb816245426d8ba21bddb2351f4d4d0301e74

packages/slate-commons/src/commands/isCursorInEmptyParagraph.ts
e1himself: In the worst case you can always refer to function arguments via Typescript `Parameters` type:

```
type Options = Parameters<isEmptyParagraphElement>[2]; 
```

aspirisen: @e1himself we could not face it, but the consumers could, like here https://github.com/DefinitelyTyped/DefinitelyTyped/pull/57710

Also there can be potential issues with naming as Options pretty general name and when you will need to export it probably you would need to rename the interface as well.

aspirisen: @e1himself yes, but it will not work with generics

e1himself: > Also there can be potential issues with naming as Options pretty general name and when you will need to export it probably you would need to rename the interface as well.

IDK, I see what you mean. But I hate making the code difficult to read without having a really good reason. The ones you named I classify as "potential" -- meaning some people may face it in the future, or not. So I'd rather keep the code simpler until there's a real reason to complicate it.

packages/slate-commons/src/commands/isEmptyParagraphElement.ts
e1himself: > imho i am not sure about "ignoreWhitespace" name
i.e. in this example "Hello world" has whitespace in the middle, but it will not be ignored

After thinking over night, I think `ignoreWhitespace` could still do the job, as it has no effect on your `Hello world` counter-example whether we *trim* whitespace or ignore all of them (`.replace(/s/g, '')`) -- the string is still not empty (`Helloworld`). So the difference is effectively indistinguishable for the use-case :shrug: :) 

But let's keep it `trim`. Makes more sense to have a shorter name.

PR #92, opened 2022-02-03
WIP: [MT-4645] Implement common UI components for the new menu toolbars

aspirisen:


packages/slate-editor/src/components/ChooseGroup/ChooseGroup.tsx
```diff
@@ -0,0 +1,58 @@
...
+                    <ChooseOption
+                        key={o.value}
+                        name={props.name}
+                        type={props.type}
+                        option={o}
+                        allSelected={allSelected}
+                        onChange={props.onChange}
+                        selected={props.selected}
+                    />
+                ),
```
e1himself, line 47: Hm... Aren't these two branches identical?

packages/slate-editor/src/components/Button/Button.tsx
```diff
@@ -0,0 +1,54 @@
...
+            background-color: transparent;
+            border: 1px solid transparent;
+        `}
+
+    ${(props) =>
+        props.fullWidth &&
+        css`
+            flex-grow: 1;
+        `}
+`;
```
e1himself, line 49: I must say that `styled-components` is impressive in that they've built a styling framework and a whole new knowledge domain. It's clever.

What I don't like is that we cannot make it work alongside with the existing styling approach -- they have zero compatibility. We can't even reuse the same variables in boths systems.

So, with all the attractiveness, I don't think it is a good thing for the project on the current stage to switch styling to `styled-components`. I'd rather consider switching to CSS Modules, as it's more like an "evolutionary" improvement comparing to what we have now, in contrast styled components will cause a "revolution", which I don't really deal with right now.

@aspirisen Sorry about this, but I truly believe it's the right choice in the current situation. :raised_hands:

packages/slate-editor/src/components/Toolbox/Toolbox.stories.tsx
```diff
@@ -0,0 +1,123 @@
...
+                        fill={props.isActive ? '#F9CA7B' : 'white'}
+                    />
+                    <path
+                        opacity="0.4"
+                        fillRule="evenodd"
+                        clipRule="evenodd"
+                        d="M42 37H22V11H42V37Z"
+                        fill="white"
+                    />
+                </svg>
```
e1himself, line 43: Let's extract this into `icons` folder as an external SVG file.

packages/slate-editor/src/components/ChooseGroup/ChooseGroup.tsx
```diff
@@ -0,0 +1,58 @@
...
+interface ChooseGroupCommon<T extends string> {
+    name: string;
+    options: ChooseGroupOption<T>[];
+    columns?: number;
+}
+
+type ChooseGroupProps<T extends string> = ChooseGroupCommon<T> &
+    (CheckboxChoose<T> | RadioChoose<T>);
+
+export function ChooseGroup<T extends string>(props: ChooseGroupProps<T>) {
```
e1himself, line 16: Hm... I don't remember ever seeing a component like this named `ChooseGroup`. Sounds awkward.

Can we check how other frameworks typically call these components?

e1himself (commented):
Good job @aspirisen :+1: 

After thinking hard about whether we should start using styled components in this project, I've came to a conclusion that it brings more problems than benefits. Sorry.

packages/slate-editor/src/components/Button/Button.tsx
aspirisen: It is possible to have the same variables, we can create a ts file that create theme values and then generate scss file with the same values. The advantage of styled-components is improved development experience. You have styles in the same file as the component, so you do not need to switch between files. As the styled components are inside TS file we can have find reference, and when the styles are no longer used TS will tell us about unused components. We would have easier refactoring when we would split components into different files we would explicitly know which styles should be also moved. Is scss we have complex classnames with concatenation and you do not have any ide support to use the class names inside react components. Classnames becomes pretty long and and the JSX starts harder to read. Also classnames becomes pretty complex when you have conditions.

In general i do not see too much difficulty in moving to styled-components as it is just css. The components that are already in scss will continue to work, and when we will change smth in the components we could change it to styled-components step by step.

By the way, plate uses styled-components as well

aspirisen: > True for standard SCSS. But not the case for CSS modules.

Pretty much the same would with css modules as well, as you have element and modifiers inside the element, in jsx you have to use this long name, i.e. the element name + element--modifier + element modifier2

aspirisen: I just looked through the components and it looks like there is not so many complicated scss files, the most complicated is Menu component styles, the others are simpler

e1himself: I can feel you are unhappy about reverting the styled components approach, but I still believe that it's better to not adopt it to this project at this current moment in time. 

Let me know if you could use my help on rewriting these components to SCSS.

digitalbase:
Hey, this was merged but I don't see a conclusion with regards to the convo around scss. Could not find it in linear either.

@aspirisen ?

e1himself:
@digitalbase we've come to an agreement that CSS Modules is the way to go.  I've written a summary here: https://linear.app/prezly/issue/MT-4676/add-css-modules-support-for-slate#comment-99037f02

This PR is (confusingly) marked as merged because it is technically a subset of #94, in which styled components were removed and CSS Modules system introduced on top.

PR #93, opened 2022-02-05
Widen the peer dependency for React

moises-marquez:
The peer dependency for React is too restrictive


e1himself:
Thanks for the pull request @moises-marquez :heart: 

I've done the same patch for the rest of sub-packages, so we have consistent requirements in all of them. 

This change will be released as a tagged version this week.

PR #101, opened 2022-02-10
[MT-4662] New components menus: Phase 1

aspirisen:


packages/slate-editor/src/components/Toolbox/Toolbox.stories.tsx
```diff
@@ -0,0 +1,67 @@
...
+                        <Toggle>Open in new tab</Toggle>
+                    </VStack>
+                </Toolbox.Section>
+
+                <Toolbox.Footer>
+                    <Button variant="clear" Icon={Delete} fullWidth>
+                        Remove web bookmark
+                    </Button>
+                </Toolbox.Footer>
+            </Toolbox.Panel>
```
e1himself, line 64: I did an extensive QA on the final look & feel here and noticed a few differences. Tried to outline them all here:
image

Smaller text (toggles labels, options labels, secondary buttons) all have font-size 13px in the designs.

Here's the design for the toggle "off" state: https://www.figma.com/file/KD30VWNjFh6mVWkg6Taqhe/06-%E2%80%93-Editor?node-id=1955%3A16012

Also see this: MT-4692

e1himself (changes requested):
Good job @aspirisen :+1: 

Please see my extensive QA report below :sweat_smile:

packages/slate-editor/src/components/Stack/Stack.tsx
```diff
@@ -0,0 +1,26 @@
...
+    spacing: StackSizes;
+    direction?: 'column' | 'row';
+}
+
+export function Stack(props: React.PropsWithChildren<StackProps>) {
+    return (
+        <div
+            className={classNames(
+                styles.stack,
+                styles[`--${props.direction ?? 'row'}-${props.spacing}`],
```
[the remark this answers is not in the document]
e1himself: > There more straightforward solution is to have a lot of "if"s with all combinations of the direction and spacing

That would be my preference. It is straightforward (which is good) :) 

But I don't want to push this. Just wondering whether you think that's worth addressing.

packages/slate-editor/src/components/OptionsGroup/Option.tsx
```diff
@@ -0,0 +1,53 @@
...
+    option: OptionsGroupOption<T>;
+    allSelected: Set<T | undefined>;
+};
+
+export function ChooseOption<T extends string>(props: ChooseOptionProps<T>) {
+    const onChange = React.useCallback(() => {
+        props.onChange?.(props.option.value);
+    }, [props.option, props.onChange]);
+
+    const isActive = props.allSelected.has(props.option.value);
```
e1himself, line 32: I see no reason to pass whole `allSelected` set from above if we only need to check if it's included.
A much more expected component API would be to only ask for `isSelected` prop from outside.
Would you agree?

packages/slate-editor/src/components/OptionsGroup/OptionsGroup.tsx
```diff
@@ -0,0 +1,38 @@
...
+interface OptionsGroupCommon<T extends string> {
+    name: string;
+    options: OptionsGroupOption<T>[];
+    columns?: number;
+}
+
+type OptionsGroupProps<T extends string> = OptionsGroupCommon<T> & RadioChoose<T>;
+
+export function OptionsGroup<T extends string>(props: OptionsGroupProps<T>) {
+    const allSelected = React.useMemo(() => new Set([props.selected]), [props.selected]);
```
e1himself, line 16: Because our `OptionsGroups` is always a radio group, the controlled state should only be one `value` prop, and not an array/set. Right?

packages/slate-editor/src/components/OptionsGroup/Option.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+export interface OptionsGroupOption<T extends string> {
+    label: string;
+    value: T;
+    Icon?: React.ComponentType<React.SVGProps<SVGSVGElement> & { isActive: boolean }>;
+}
+
+interface Option<T> {
+    selected?: T;
+    onChange?: (value: T | undefined) => void;
```
e1himself, line 14: `undefined` should not be possible for `onChange` handler

packages/slate-editor/src/components/OptionsGroup/Option.tsx
```diff
@@ -0,0 +1,48 @@
...
+    label: string;
+    value: T;
+    Icon?: React.ComponentType<React.SVGProps<SVGSVGElement> & { isActive: boolean }>;
+    disabled?: boolean;
+}
+
+export interface ChooseOptionProps<T extends string> {
+    name: string;
+    option: OptionsGroupOption<T>;
+    selected?: T;
```
e1himself, line 18: nit: IMO the least surprizing API for these kind of components would be to have `selected` as boolean (and lift `selected={option.value === selectedValue}` one level up). Similar to `<input type="radio">` HTML props interface.

packages/slate-editor/src/components/OptionsGroup/OptionsGroup.module.scss
```diff
@@ -0,0 +1,51 @@
...
+
+.label {
+    $border-size: 4px;
+
+    padding-top: 4px;
+    cursor: pointer;
+    display: inline-flex;
+    flex-flow: column;
+    align-items: center;
+    font-size: $font-size-small;
```
e1himself, line 22:
```suggestion
    font-size: $font-size-smaller;
    font-weight: 600;
```

packages/slate-editor/src/components/OptionsGroup/OptionsGroup.tsx
```diff
@@ -0,0 +1,35 @@
...
+import styles from './OptionsGroup.module.scss';
+
+interface OptionsGroupCommon<T extends string> {
+    name: string;
+    options: OptionsGroupOption<T>[];
+    columns?: number;
+}
+
+type OptionsGroupProps<T extends string> = OptionsGroupCommon<T> &
+    Omit<ChooseOptionProps<T>, 'option'>;
```
e1himself, line 14: nit: I find it difficult to trace that `OptionsGroupProps` extends `OptionsGroupCommon` and `ChooseOptionProps`. Let's just inline those props here. I see no harm in slight duplication here.

packages/slate-editor/src/components/Toggle/Toggle.module.scss
```diff
@@ -0,0 +1,58 @@
+@import "styles/variables";
+
+.toggle {
+    color: $white;
+    font-size: $font-size-small;
+    font-weight: normal;
```
e1himself, line 6:
```suggestion
    font-weight: 500;
```

packages/slate-editor/src/components/OptionsGroup/Option.tsx
```diff
@@ -0,0 +1,48 @@
...
+    onChange: (value: T | undefined) => void;
```
e1himself, line 19: Why do we have `undefined` in `onChange` here? Under which conditions it can produce `onChange(undefined)`?

packages/slate-editor/src/components/OptionsGroup/OptionsGroup.tsx
```diff
@@ -0,0 +1,35 @@
+import * as React from 'react';
+
+import type { OptionsGroupOption, ChooseOptionProps } from './Option';
+import { ChooseOption } from './Option';
+import styles from './OptionsGroup.module.scss';
+
+interface OptionsGroupCommon<T extends string> {
+    name: string;
+    options: OptionsGroupOption<T>[];
+    columns?: number;
```
e1himself, line 10: What do you think about adding a top-level `disabled` prop here to disable all options at once? Would be more convenient, and also we can keep the `options` array as a const and only dynamically calculate top-level `disabled` prop on the whole `OptionsGroup` parent.

packages/slate-editor/src/components/OptionsGroup/OptionsGroup.stories.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+    return (
+        <div style={{ background: 'rgba(27, 27, 54, 0.96)', padding: 20 }}>
+            <OptionsGroup
+                name="card-layout"
+                options={cardLayoutOptions}
+                selected={cardLayout}
+                onChange={setCardLayout}
+                columns={3}
+            />
```
e1himself, line 43: Please carefully go over fonts in the designs and compare it to the implementation. I still see several differences:

image

packages/slate-editor/package.json
```diff
@@ -47,7 +47,7 @@
         "build:esm": "gulp build:esm",
         "build:cjs": "gulp build:cjs",
         "build:sass": "gulp build:sass",
-        "watch": "concurrently --kill-others --names 'd.ts,esm ,cjs ,sass' 'npm run watch:types' 'npm run watch:esm' 'npm run watch:cjs' 'npm run watch:sass'",
+        "watch": "concurrently --restart-tries 5 --kill-others --names 'd.ts,esm ,cjs ,sass' 'npm run watch:types' 'npm run watch:esm' 'npm run watch:cjs' 'npm run watch:sass'",
```
e1himself, line 50: :+1:

.eslintrc.cjs
```diff
@@ -30,6 +30,7 @@ module.exports = {
         }
     },
     "rules": {
+        "no-fallthrough": "off",
```
e1himself, line 33: :+1:

e1himself (changes requested):
Great job @aspirisen :+1: 

I love that all the components have been implemented with a11y in mind (having real HTML inputs internally). Kudos for that :1st_place_medal: 

There are a few visual differences still remaining. 

Plus I have a few nitpicking comments. Please see below.

e1himself (approved):
Good job Dmitri! :+1: 

My heart is filled with joy :smiley:

PR #102, opened 2022-02-11
[MT-4646] Feature - New component toolbox popover container

e1himself:
Peek 2022-02-14 13-55


packages/slate-editor/src/components/EditorBlock/EditorBlock.tsx
```diff
@@ -0,0 +1,73 @@
...
+                {children}
+            </div>
+
+            {/* We have to render children or Slate will fail when trying to find the node. */}
+            {slateInternalsChildren}
+        </div>
+    );
+});
+
+EditorBlock.displayName = 'EditorBlock';
```
aspirisen, line 73: q: Maybe we could use babel plugin for that https://www.npmjs.com/package/babel-plugin-react-anonymous-display-name ?

aspirisen (approved):

packages/slate-editor/src/components/EditorBlock/EditorBlock.tsx
e1himself: Yeah, why not! :)

PR #103, opened 2022-02-14
[MT-4618] Feature - Web Bookmark menu

e1himself:
Based on #101 and #102

Peek 2022-02-14 16-56


aspirisen:
Happy So Cool GIF by Simon Super Rabbit

PR #104, opened 2022-02-14
[MT-4617] Feature - New focus state colors

e1himself:
- Update focus outline to match the new designs: make it yellow and thicker (6px vs former 4px)
- Expand Divider block hit area and focus outline to take more visual space (keeping the actual spacing between blocks the same)
- Drop floating toolbar from Divider block (with only "delete" button)
  See https://linear.app/prezly/issue/MT-4613/rework-editor-blocks-menus-and-uis#comment-77fe5251

Peek 2022-02-14 17-35


aspirisen:
3ohzdIuqJoo8QdKlnW

aspirisen (approved):

PR #105, opened 2022-02-16
[MT-4673] Fix - Pressing UP key doesnt focus some void elements

e1himself:
Based on #104 

- Fix selection of Divider and Bookmark blocks with keyboard UP/DOWN navigation
- Fix Bookmark block selection state to match the designs (no rounded borders; no gray border)
- Refactor `EditorBlock` to pass Slate internal children as `children`

### DIFF
--- a/packages/slate-editor/src/components/EditorBlock/EditorBlock.module.scss
+++ b/packages/slate-editor/src/components/EditorBlock/EditorBlock.module.scss
@@ -4,11 +4,6 @@
 .block {
     @include editor-v4-block($void: false);
     position: relative;
-    white-space: normal;
-
-    &.void {
-        @include editor-v4-void-element;
-    }
 
     &.selected {
         @include editor-v4-block-active;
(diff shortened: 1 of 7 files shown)


packages/slate-editor/src/components/EditorBlock/EditorBlock.module.scss
```diff
@@ -4,11 +4,6 @@
 .block {
     @include editor-v4-block($void: false);
     position: relative;
-    white-space: normal;
```
e1himself, line 7: This was causing the problem. 

And also setting `user-select: none` for `.void` elements too.

Big kudos to @aspirisen for helping finding it. :trophy:

packages/slate-editor/src/styles/shared.scss
```diff
@@ -1,3 +1,4 @@
+$editor-v4-block-focus-color: $yellow-300;
```
aspirisen, line 1: Looks like missing "variables" import

aspirisen (approved):

PR #107, opened 2022-02-16
[MT-4552] Feature - HTML blocks support

e1himself:
- Add `html` node type to the model
- Add a simple block renderer to the editor


e1himself:
Pinging @aspirisen just as FYI. There's not much to review here :) Will merge without a review.

PR #118, opened 2022-02-25
[DEV-4541] Story Embed rich content block

pestaa:
Companion PR:

* https://github.com/prezly/prezly/pull/10722


packages/slate-editor/src/modules/editor-v4-story-embed/components/FloatingStoryEmbedInput/FloatingStoryEmbedInput.scss
```diff
@@ -0,0 +1,30 @@
+@use "sass:math";
```
aspirisen, line 1: we now use css modules, take a look on packages/slate-editor/src/components/Button for usage example

PR #123, opened 2022-03-04
[MT-4616] Design fixes for link menu

aspirisen:
https://linear.app/prezly/issue/MT-4865/input-focused-state-styling-dont-match-designs
https://linear.app/prezly/issue/MT-4866/arrow-on-the-top-of-the-overlay-is-missing
https://linear.app/prezly/issue/MT-4867/highlight-the-text-you-are-linking


e1himself (commented):
Hey! 

Good job! 

I've found a few problems with the current implementation.

1. Applying selection marks to the document state is not the best idea. It can produce leftovers in the document value object with the marks not supported by the backend. I had this case locally after clicking link button and then dismissing the link menu after that. 
    I believe I've fixed it in #124 by refactoring selection rendering to using `decorate` functionality.

2. Link menu positioning does not match the design. Fixed in #125 

3. There are glitches with the link menu staying open after I click somewhere else in the document. Didn't find a way to fix it.

4. Invalid state for the URL input has weird sizing glitch (Firefox only?).  The right edge of the input moves to the right, when the :warning: icon is displayed.

packages/slate-editor/src/modules/editor-v4-rich-formatting-menu/RichFormattingMenu.tsx
```diff
@@ -138,12 +139,21 @@ export const RichFormattingMenu: FunctionComponent<Props> = ({
         Transforms.select(editor, selection);
     }
 
+    React.useEffect(() => {
+        Editor.addMark(editor, MarkType.SELECTION, true);
+
+        return () => {
+            Editor.removeMark(editor, MarkType.SELECTION);
+        };
+    }, [linkRange?.current]);
```
e1himself, line 148: Applying selection marks to the document state is not the best idea. It can produce leftovers in the document value object with the marks not supported by the backend. I had this case locally after clicking link button and then dismissing the link menu after that. Selection marks were kept in the document even after I selection something else.
I believe I've fixed it in https://github.com/prezly/slate/pull/124 by refactoring selection rendering to using decorate functionality.

aspirisen:
@e1himself thank you, regarding shifting the right edge - when there is no error we don't add padding, when there is error we add padding to show the icon. I am not sure if we need to always save the space for the icon because by that we will hide part of the text and in general it could look strange if you see that there is space for more text but it is not visible

e1himself:
@aspirisen 

>  regarding shifting the right edge - when there is no error we don't add padding, when there is error we add padding to show the icon. I am not sure if we need to always save the space for the icon because by that we will hide part of the text and in general it could look strange if you see that there is space for more text but it is not visible

I don't understand. It clearly looks broken -- when the icon appears, the input grows bigger. This can't be right.

aspirisen:
@e1himself ah, that's strange, probably firefox issue, i will check

aspirisen:
@e1himself should be fixed now

e1himself:
> @e1himself should be fixed now

@aspirisen Looks good now. Thanks! :+1:

e1himself (approved):
:+1:

PR #124, opened 2022-03-06
[MT-4616] Refactor - Selection mark to runtime decorations

e1himself:
- Refactor Extension `decorate` property to be a factory (not an implementation). 
  This allows extensions to define their decorators depending on the `editor` instance.
  This also matches Plugin `decorate` property from Plate: 
  https://github.com/udecode/plate/blob/deeab96a077349ac607eaeffd7bb6f4f73d60832/packages/core/src/types/plugins/Decorate.ts#L5-L13 

- Rewrite text `selection` mark from being a data model to a runtime decorator.
  See https://docs.slatejs.org/concepts/09-rendering#decorations


packages/slate-editor/src/modules/editor-v4-rich-formatting/components/Text.tsx
```diff
@@ -26,13 +24,5 @@ export function Text({ attributes, children, leaf }: RenderLeafProps) {
...
 
-    if (leaf[MarkType.SELECTION]) {
-        children = (
-            <span {...attributes} className={styled.selection}>
-                {children}
-            </span>
-        );
-    }
-
-    return <span {...attributes}>{children}</span>;
```
e1himself, line 37: `{...attributes}` is unnecessary here, as it is already rendered in `renderLeafExtensions`: https://github.com/prezly/slate/blob/824f64d237623935fb0927ad33055d3b463913c5/packages/slate-commons/src/lib/renderLeafExtensions.tsx#L22

aspirisen (approved):

aspirisen:
Dog Thank You GIF by MOODMAN (1)

PR #127, opened 2022-03-08
[MT-4866] Fix Link Menu arrow positionioning

e1himself:
https://github.com/prezly/slate/pull/126#issue-1162705401

> There is still issue with arrow position when the target is huge, it is placed in the middle, but according to the designs it should be always on the left, thoughts?


aspirisen:
Rock On Reaction GIF
Thank you :)

aspirisen (approved):

PR #131, opened 2022-03-11
[DEV-5017] Assign value to editor children

aspirisen:
From https://github.com/ianstormtaylor/slate/pull/4540#issuecomment-951380551


packages/slate-editor/src/modules/editor-v4/EditorV4.tsx
```diff
@@ -140,6 +140,10 @@ const EditorV4: FunctionComponent<EditorV4Props> = (props) => {
         plugins,
     });
 
+    React.useEffect(() => {
```
OskarGasiorowski, line 143: I think we should be consistent and use `useEffect` instead `React.useEffect` 😄

OskarGasiorowski (approved):

PR #135, opened 2022-03-16
Refactor - Deprecate ImageNode `width_factor` property

e1himself:
The attribute does not seem to be used at all. I've checked the database and there are only a few image nodes having `width_factor` set to anything else except `100%`. All used in a few fairly old stories dating to 2018.

### DIFF
--- a/packages/slate-types/src/nodes/ImageNode.ts
+++ b/packages/slate-types/src/nodes/ImageNode.ts
@@ -20,10 +20,6 @@ export interface ImageNode extends ElementNode {
     layout: ImageLayout;
     /** matches this regexp: /^\d+(\.\d+)?%$/ */
     width: string;
-    /**
-     * @deprecated No longer used. Will be dropped in the future.
-     */
-    width_factor: '100%';
 }
 
 export function isImageNode(value: any): value is ImageNode {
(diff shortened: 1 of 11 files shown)


packages/slate-types/src/nodes/ImageNode.ts
```diff
@@ -20,8 +20,10 @@ export interface ImageNode extends ElementNode {
     layout: ImageLayout;
     /** matches this regexp: /^\d+(\.\d+)?%$/ */
     width: string;
-    /** matches this regexp: /^\d+(\.\d+)?%$/ */
-    width_factor: string;
+    /**
+     * @deprecated No longer used. Will be dropped in the future.
+     */
+    width_factor: '100%';
```
aspirisen, line 26: Why do we need to keep the prop in types?

e1himself: I was hesitating to do so, as it would be a BC breaking change. But after thinking this through I couldn't think of any problems caused by dropping it. So let it be. a59fc3cd

Thanks.

aspirisen:
radio remains GIF
👍

packages/slate-types/src/nodes/ImageNode.ts
pestaa: I have a suspicion that dropping the property caused https://linear.app/prezly/issue/DEV-5142/customers-send-campaign-button-is-greyed-out-and-is-unable-to-send

Because the local state does not match the state coming from backend after a save -- so it keeps believing there are unsaved changes.

Taking care of it now. :)

PR #136, opened 2022-03-28
[MT-4873] Added option to turn bold formatting on Heading One and Two

OskarGasiorowski:


e1himself:
Code-wise it looks good. But I think that the chosen solution will bite us in the ass.

Please see my comment https://linear.app/prezly/issue/MT-4873/bolding-text-appears-the-opposite-in-the-story-editor#comment-dce7fb95

e1himself:
@OskarGasiorowski should we close this PR?

PR #139, opened 2022-04-01
[MT-4881] fixed placeholder issue with mentions

OskarGasiorowski:


packages/slate-editor/src/modules/editor-v4-floating-add-menu/lib/shouldShowMenuButton.ts
```diff
@@ -17,6 +17,9 @@ export function shouldShowMenuButton(editor: Editor): boolean {
     }
 
     const text = Node.string(currentNode);
+    const hasNoPlaceholder = currentNode.children.every(
+        (child) => !('type' in child && child.type === PLACEHOLDER_NODE_TYPE),
```
aspirisen, line 21: I think it is better to use Node.nodes(currentNode) to get children and use isPlaceholderNode to detect if a node is a placeholder

Also i think it is better to check that the first node is not a placeholder, not that there are some placeholders in the text

aspirisen: You can use  `Array.from(Node.nodes(currentNode))`
> For some reason, the first node is empty text, so I can't be sure in which position it can be.

As i understand the issue is when the placeholder is first node in paragraph, if you would have space and the placeholder as i understand there will be no issues

OskarGasiorowski: So, just check always second place for placeholder?

aspirisen: Ok, i got what you mean, i think you can leave checking for all children. But i would reverse logic, i would check if line has placeholders and used currentNode.children.some

OskarGasiorowski: done

packages/slate-editor/src/modules/editor-v4-floating-add-menu/lib/shouldShowMenuButton.ts
```diff
@@ -17,6 +17,9 @@ export function shouldShowMenuButton(editor: Editor): boolean {
     }
 
     const text = Node.string(currentNode);
+    const hasPlaceholder = currentNode.children.some(
+        (child) => 'type' in child && child.type === PLACEHOLDER_NODE_TYPE,
```
aspirisen, line 21: Could you use Array.from(Node.nodes(currentNode)) and isPlaceholderNode

packages/slate-editor/src/modules/editor-v4-floating-add-menu/lib/shouldShowMenuButton.ts
```diff
@@ -17,6 +17,9 @@ export function shouldShowMenuButton(editor: Editor): boolean {
     }
 
     const text = Node.string(currentNode);
+    const hasPlaceholder = currentNode.children.some(
```
e1himself, line 20: I suggest generalizing the check and instead of looking for placeholder nodes, consider a paragraph as non-empty if there are any void children (like placeholder).

From practical point of view, the original code can be modified to:

```diff
- if (!isParagraphNode(currentNode)) {
+ if (!isParagraphNode(currentNode) || hasVoidElements(editor, currentNode)) {
    return false;
  }
```

`hasVoidElements()` is already implemented in *isEmptyParagraphElement.ts* (so you'll need to extract it to a separate exported function in the `commands` folder).

OskarGasiorowski: Hmm. I don't see anything like that: 
Screenshot 2022-04-05 at 15 34 06

e1himself: https://github.com/prezly/slate/blob/master/packages/slate-commons/src/commands/isEmptyParagraphElement.ts

OskarGasiorowski: done

packages/slate-editor/src/modules/editor-v4-floating-add-menu/lib/shouldShowMenuButton.ts
OskarGasiorowski: I went with @e1himself solution

aspirisen (approved):

e1himself (approved):
Nice! :+1:

PR #142, opened 2022-04-04
[MT-4897] Add story card UI

aspirisen:


packages/slate-editor/src/modules/editor-v4-story-bookmark/components/StoryBookmarkElement/StoryBookmarkElement.tsx
```diff
@@ -28,6 +31,10 @@ export function StoryBookmarkElement({ attributes, children, element, params }:
         loadStory();
     }, [loadStory]);
 
+    if (error) {
+        console.log(error);
```
aspirisen, line 35: @e1himself do you know already made solutions for handling errors?

packages/slate-editor/src/modules/editor-v4-components/BookmarkCard/BookmarkCard.module.scss
```diff
@@ -128,12 +129,12 @@
     }
 }
 
-.providerIcon {
+.provider-icon {
```
e1himself, line 132: To be honest I don't see a point in these renaimings. What's the reason for making our lives harder and the code less readable? 

`styles.providerIcon` vs `styles['provider-icon']` ?

packages/slate-editor/src/modules/editor-v4-components/BookmarkCard/BookmarkCard.stories.tsx
```diff
@@ -0,0 +1,74 @@
...
+                    title={info.title}
+                    description={info.description}
+                >
+                    <BookmarkCard.Provider
+                        showUrl={props.showUrl}
+                        url={info.url}
+                        providerName={info.provider_name}
+                    />
+                </BookmarkCard.Details>
+            </BookmarkCard.Container>
```
e1himself, line 45: Did you consider making the `BookmarkCard` component a black box where everything is passed inside as properties? (in contrast to building a card with sub-components)

IMO it's just easier to use this way and less code duplication.

```tsx
<BookmarkCard
   url={...}
   title={...}
   description={...}
   providerName={...}
   thumbnailUrl={...}
   thumbnailWidth={...}
   thumbnailHeight={...}
/>
```

packages/slate-editor/src/modules/editor-v4-story-bookmark/components/StoryBookmarkElement/StoryBookmarkElement.tsx
e1himself: https://github.com/prezly/slate/blob/b9758a8db1016335e59f6beebfb7603a0017ab62/packages/slate-editor/src/modules/editor-v4-embed/lib/useFloatingEmbedInput.ts#L89-L93

packages/slate-editor/src/modules/editor-v4-components/BookmarkCard/BookmarkCard.module.scss
aspirisen: To keep consistency in css naming, everywhere in the app we use this approach as i can see

packages/slate-editor/src/modules/editor-v4-components/BookmarkCard/BookmarkCard.stories.tsx
aspirisen: There can be conditions in rendering different parts of the card, or if you need to wrap some nodes with other elements, or if you need to add some custom elements inside the card. Thus you have to add new props to cover all these needs

packages/slate-editor/src/modules/editor-v4-components/BookmarkCard/BookmarkCard.module.scss
e1himself: > To keep consistency in css naming, everywhere in the app we use this approach as i can see

But it did make sense before CSS modules, now it doesn't.

packages/slate-editor/src/modules/editor-v4-components/BookmarkCard/BookmarkCard.stories.tsx
e1himself: Are you saying that this *might possibly* be needed in a future use-case that does not exist yet? If so, I'd suggest to keep things simple now and complicate them only when there is a need for that.

aspirisen: We already have conditions in JSX
personally i would go with composition if it is possible. Thus in JSX you see where the props are bypassed rather than see a component with a bunch of props with layouting like condition to render this or that component

e1himself: I don't share your point of view. But alright, I won't continue this discussion. Let's do it this way, with composition.

e1himself (approved):
:+1:

packages/slate-editor/src/modules/editor-v4-story-bookmark/components/StoryBookmarkElement/StoryBookmarkBlock.tsx
```diff
@@ -0,0 +1,65 @@
...
+    const isEmpty =
+        !showThumbnail &&
+        utils.isEmptyText(story.oembed.title) &&
+        utils.isEmptyText(story.oembed.description);
+
+    const actualLayout = !showThumbnail
+        ? StoryBookmarkLayout.HORIZONTAL
+        : isSmallViewport
+        ? StoryBookmarkLayout.VERTICAL
+        : element.layout;
```
OskarGasiorowski, line 33: Maybe you should consider something like that? `smallOrHorizontalViewport` is just a fast idea, if you have a better name for that feel free to change :)

```suggestion
    const smallOrHorizontalViewport = !showThumbnail
        ? StoryBookmarkLayout.HORIZONTAL
        : isSmallViewport;
    const actualLayout = smallOrHorizontalViewport
        ? StoryBookmarkLayout.VERTICAL
        : element.layout;
```

Issue #132, opened 2022-03-11
Module "slate" declare conflict


tilemoon:
<img width="777" alt="image" src="https://user-images.githubusercontent.com/37775224/157862010-19674e55-f424-4697-9a35-76f6a5deeee7.png">

When i import something from @prezy/slate-lists, there will be an error below.
<img width="858" alt="image" src="https://user-images.githubusercontent.com/37775224/157862334-19f90db7-5cf6-4c04-87df-30bbc94eeadb.png">

wildf1r3:
Hi @tilemoon this seems to be the same problem as in issue #49 , im also waiting for an answer from the author on this regard, if you find any suitable solution other  than just forking and fixing it yourself, please let me know!

Thanks a bunch!

PR #148, opened 2022-04-15
Refactor - Slate Lists package

### LINKED ISSUE #49: slate-lists: Cannot customize types
Because slate custom types are declared here:
https://github.com/prezly/slate/blob/master/packages/slate-lists/src/index.ts#L6-L12

I'm having trouble creating my own custom types:

```ts
Subsequent property declarations must have the same type.  Property 'Text' must be of type 'TextNode', but here has type 'CustomText'.ts(2717)
index.d.ts(9, 9): 'Text' was also declared here.
(property) CustomTypes.Text: TextNode
```

image

What's the right way to customize types when using this plugin? Maybe that module declaration shouldn't be in the package?

e1himself:
Fixes #49 and #132 

- Refactor `slate-commons` to not depend on global slate custom types declaration
- Refactor `slate-lists` to not depend on `slate-types` or global slate custom types declaration
- Rework `slate-lists` functionality for simplicity (extend `Editor` instance with necessary options instead of carrying over `ListOptions` object with every call)
- Move lists-related `onKeyDown` handler into `slate-lists` package, so package consumers can easily reuse this functionality

# Usage

See example initialization at [withListsFormatting.ts](https://github.com/prezly/slate/blob/356fdf8c3461f9d98a812efcf8b76b8f779666b8/packages/slate-editor/src/modules/editor-v4-rich-formatting/withListsFormatting.ts).

```ts
const SCHEMA: ListsSchema = {
    isConvertibleToListTextNode(node): boolean {
        return isParagraphNode(node) || isHeadingNode(node) || isQuoteNode(node);
    },
    isDefaultTextNode: isParagraphNode,
    isListNode(node, type?) {
        if (type === ListType.ORDERED) {
            return isListNode(node, NUMBERED_LIST_NODE_TYPE);
        }
        if (type === ListType.UNORDERED) {
            return isListNode(node, BULLETED_LIST_NODE_TYPE);
        }
        return isListNode(node);
    },
    isListItemNode,
    isListItemTextNode,
    createDefaultTextNode({ children } = {}) {
        return createParagraph({ children });
    },
    createListNode(type: ListType = ListType.UNORDERED, { children } = {}) {
        return {
            type: type === ListType.ORDERED ? NUMBERED_LIST_NODE_TYPE : BULLETED_LIST_NODE_TYPE,
            children: children ?? [this.createListItemNode()],
        };
    },
    createListItemNode({ children } = {}) {
        return { 
            type: LIST_ITEM_NODE_TYPE,
            children: children ?? [this.createListItemTextNode()], 
        };
    },
    createListItemTextNode({ children } = {}) {
        return { 
           type: LIST_ITEM_TEXT_NODE_TYPE,
           children: children ?? [{ text: '' }], 
        };
    },
};

const editor = withListsReact(withLists(SCHEMA)(withReact(createEditor())));
```

# Hotkeys

With these changes, we now export `onKeyDown` handler from `slate-lists`, which can be used with your Editor instance to implement keyboard interactions with lists: `Tab` / `Shift+Tab` to increase/decrease list nesting, etc.

```tsx
import { onKeyDown } from '@prezly/slate-lists';

<Editable
   // ...
   onKeyDown={function (event) {
      onKeyDown(editor, event);
   })
/>
```


wildf1r3 (approved):
These changes feel good, i like the new ListSchema interface and the ability to freely define given functions, as this feels like the way slate is intended to be used/slate plugins are intended to be written

PR #149, opened 2022-04-20
[MT-4878] Fix - Page jumps when interacting with embedded story

OskarGasiorowski:


packages/slate-editor/src/modules/editor-v4/lib/useCursorInView.ts
```diff
@@ -34,7 +34,10 @@ function ensureCursorInView(
 
     const [currentNode] = EditorCommands.getCurrentNodeEntry(editor) || [];
 
-    if (Editor.isBlock(editor, currentNode) && Editor.isVoid(editor, currentNode)) {
+    if (
+        Editor.isBlock(editor, currentNode) ||
+        (Editor.isBlock(editor, currentNode) && Editor.isVoid(editor, currentNode))
```
e1himself, line 39: I don't understand this change. 

`Editor.isBlock(editor, currentNode) || Editor.isBlock(editor, currentNode) && Editor.isVoid(editor, currentNode)` is equivalent to just `Editor.isBlock(editor, currentNode)`.

aspirisen (approved):

packages/slate-editor/src/lib/scrollIntoView.ts
```diff
@@ -44,11 +44,6 @@ export function scrollIntoView(
         return;
     }
 
-    if (isChildAboveVisibleArea) {
-        const y = parent.scrollTop + elementTop - minTop;
-        scrollTo(parent, parent.scrollLeft, y);
-    } else if (isChildBelowVisibleArea) {
-        const y = parent.scrollTop + elementTop + elementHeight - parentHeight + minBottom;
-        scrollTo(parent, parent.scrollLeft, y);
-    }
```
e1himself, line 53: I believe that the change to this part of logic has introduced a major UX regression for interactions with the content. 

It now always jumps the viewport to put the cursor to the top of the viewport. That makes the editing experience suboptimal (to say the least). I didn't see any other editor out there behaving this way.


https://user-images.githubusercontent.com/370680/165104597-96a64eb5-4fd1-466c-b9d1-9ca3a76c74c9.mp4

packages/slate-editor/src/lib/scrollIntoView.ts
```diff
@@ -44,11 +44,8 @@ export function scrollIntoView(
         return;
     }
 
-    if (isChildAboveVisibleArea) {
+    if (isChildAboveVisibleArea || isChildBelowVisibleArea) {
         const y = parent.scrollTop + elementTop - minTop;
         scrollTo(parent, parent.scrollLeft, y);
-    } else if (isChildBelowVisibleArea) {
-        const y = parent.scrollTop + elementTop + elementHeight - parentHeight + minBottom;
-        scrollTo(parent, parent.scrollLeft, y);
```
e1himself, line 52: I still think the changed behavior is a step back from what it was before.

Scenarios:

1. I click on an image that is partially visible from below the viewport.

   The content jumps so that the image top edge is aligned with the viewport top edge.
   It also often happens that after I release the mouse button, it selects a portion of text. I believe this happens because I release the button outside of the clicked element, so it's treated as a range selection.

Peek 2022-04-27 14-33

2. I use the keyboard `Down` button to navigate through content. 

    Normal editor behavior is to reveal content from below the viewport as I navigate, scrolling just enough to display the focused portion. When I navigate through a paragraph, it shows the next line at the bottom of the viewport. It doesn't scroll the viewport to put the focused line at the top.

   In contrast, when I focus a block element (like a header, divider, or an image) with a keyboard, it jumps to put the element at the top edge of the editor. This makes me lose orientation. It's not clear what happens, as suddenly the screen contains whole new content (there is no visual connection to any of the previously visible content).

Peek 2022-04-27 14-34

packages/slate-editor/src/lib/scrollIntoView.ts
```diff
@@ -44,11 +44,17 @@ export function scrollIntoView(
         return;
     }
 
-    if (isChildAboveVisibleArea) {
-        const y = parent.scrollTop + elementTop - minTop;
+    const isElementHigherThanParent = elementHeight + minBottom + minTop > parentHeight;
+
+    if (isChildBelowVisibleArea && !isElementHigherThanParent) {
+        const y = parent.scrollTop + elementHeight;
```
e1himself, line 50: The math here has a problem.

Previously it was 

```ts
const y = parent.scrollTop + elementTop + elementHeight - parentHeight + minBottom;
```

and now it's just

```ts
const y = parent.scrollTop + elementHeight;
```

This leads to scroll-to following keyboard navigation not displaying blocks properly.

I'm pressing <kbd>↓</kbd> on keyboard multiple times here: 

https://user-images.githubusercontent.com/370680/168607551-30937a8d-c4b4-48cd-8224-dcb569fefd96.mp4

e1himself (changes requested):

packages/slate-editor/src/lib/scrollIntoView.ts
e1himself: Fixed by merging the code from #174

e1himself (approved):

PR #152, opened 2022-04-22
[MT-4872] Feature - New ImageNode model features

e1himself:
- Update `ImageNode` to support new `align` property
- Update `ImageNode` to support absolute sizing options in pixels
- No UI changes yet

### DIFF
--- a/packages/slate-editor/src/components/EditorBlock/Size.ts
+++ b/packages/slate-editor/src/components/EditorBlock/Size.ts
@@ -1,5 +1,7 @@
 export type Size<U extends Unit = Unit> = { value: number; unit: U };
 
+export type SizeString = `${number}%` | `${number}px`;
+
 export enum Unit {
     PIXELS = 'px',
     PERCENTS = '%',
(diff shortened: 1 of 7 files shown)


aspirisen (approved):

packages/slate-editor/src/components/EditorBlock/Size.ts
```diff
@@ -1,5 +1,7 @@
 export type Size<U extends Unit = Unit> = { value: number; unit: U };
 
+export type SizeString = `${number}%` | `${number}px`;
```
aspirisen, line 3: 👍

packages/slate-types/src/nodes/ImageNode.ts
```diff
@@ -11,16 +12,17 @@ export enum ImageLayout {
...
     file: UploadcareImageStoragePayload;
     /** empty string if no URL */
     href: string;
     layout: ImageLayout;
+    align: Alignment;
     new_tab: boolean;
-    /** matches this regexp: /^\d+(\.\d+)?%$/ */
-    width: string;
+    /** matches this regexp: /^\d+(\.\d+)?(%|px)$/ */
+    width: `${number}%` | `${number}px`;
```
OskarGasiorowski, line 25: why not as `SizeString`?

OskarGasiorowski (approved):

packages/slate-types/src/nodes/ImageNode.ts
e1himself: You are right that this is the same type, but I didn't want to couple `Size` library module (`slate-editor` package) and this generic `ImageNode` interface (`slate-types` package).

I've deduplicated the type definition by exporting the `ImageWidth` type from the `@prezly/slate-types` package.

OskarGasiorowski: oh, I haven't seen this is in two different packages. My bad :)

PR #153, opened 2022-04-26
[MT-4888] disabled button in FileAttachmentMenu on invalid filename

OskarGasiorowski:


packages/slate-editor/src/modules/editor-v4-file-attachment/components/FileAttachmentMenu.tsx
```diff
@@ -46,6 +47,14 @@ export const FileAttachmentMenu: FunctionComponent<Props> = ({
         onClose();
     };
 
+    const onFilenameChanged = useCallback(
+        function (value: string, isValid: boolean) {
+            setFilename(value);
+            setIsFilenameValid(isValid);
+        },
+        [setIsFilenameValid],
```
aspirisen, line 55: you don't need to put in deps functions that are come from react, they are always the same

aspirisen (approved):

PR #154, opened 2022-04-26
Dev env - Improve CI performance

aspirisen:
Discussion https://prezly.slack.com/archives/C029LTE0M/p1651013990025259


.github/workflows/build.yml
```diff
@@ -14,6 +14,18 @@ jobs:
     steps:
     - uses: actions/checkout@v2
 
+    - name: Cache node modules
+      uses: actions/cache@v3
+      env:
+        cache-name: cache-node-modules
+      with:
+        path: "**/node_modules"
```
e1himself, line 22: > `**/node_modules`

I wonder how this wildcard works. 

Will it cache all `node_modules` folders found in the subtree?

e1himself (approved):

.github/workflows/build.yml
aspirisen: Yes, and restores them by these paths, if some of the lock files will change (we have only one in the roor) the cache will not be applied and we will have clear npm ci on the actions

PR #155, opened 2022-04-27
[MT-4670] Fix - Placeholder should not be visible for white-space-only text blocks

OskarGasiorowski:
https://user-images.githubusercontent.com/41072665/165494467-56a0b84a-628f-46a0-b9f4-5f96fba3e999.mp4

### DIFF
--- a/packages/slate-editor/src/modules/editor-v4-floating-add-menu/FloatingAddMenu.tsx
+++ b/packages/slate-editor/src/modules/editor-v4-floating-add-menu/FloatingAddMenu.tsx
@@ -10,7 +10,7 @@ import { isHotkey } from 'is-hotkey';
 import type { KeyboardEvent, RefObject } from 'react';
 import React, { useState } from 'react';
 import type { Modifier } from 'react-popper';
-import { Transforms } from 'slate';
+import { Node, Transforms } from 'slate';
 import { useSlate } from 'slate-react';
 
 import { KeyboardKey, TooltipV2 } from '#components';
@@ -83,9 +83,9 @@ export function FloatingAddMenu<Action>({
             rememberEditorSelection();
         },
         onClose() {
+            setInput('');
             restoreEditorSelection();
             resetSelectedOption();
-            setInput('');
         },
     });
 
@@ -124,6 +124,7 @@ export function FloatingAddMenu<Action>({
     const isParagraph = isParagraphNode(currentNode);
     const isHeading1 = isHeadingNode(currentNode, HEADING_1_NODE_TYPE);
     const isHeading2 = isHeadingNode(currentNode, HEADING_2_NODE_TYPE);
+    const text = currentNode ? Node.string(currentNode) : '';
 
     return (
         <FloatingContainer.Container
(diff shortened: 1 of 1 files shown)


packages/slate-editor/src/modules/editor-v4-floating-add-menu/FloatingAddMenu.tsx
```diff
@@ -196,3 +197,7 @@ export function FloatingAddMenu<Action>({
         </FloatingContainer.Container>
     );
 }
+
+function hasOnlySpaces(text: string) {
```
OskarGasiorowski, line 201: @e1himself @aspirisen If you have better place for that let me know :)

packages/slate-editor/src/modules/editor-v4-floating-add-menu/FloatingAddMenu.tsx
```diff
@@ -164,7 +165,7 @@ export function FloatingAddMenu<Action>({
                     />
                 )}
             </TooltipV2.Tooltip>
-            {!open && (
+            {!open && !hasOnlySpaces(text) && (
```
e1himself, line 168: @OskarGasiorowski @aspirisen 

Do you guys think we should additionally disable `/` hotkey reaction when the placeholder is not visible?

Peek 2022-04-27 14-48

e1himself (approved):
Nice! :+1:

packages/slate-editor/src/modules/editor-v4-floating-add-menu/FloatingAddMenu.tsx
OskarGasiorowski: So basically `/` hotkey should work only if a line is empty?

e1himself: Maybe, but I'm not sure yet.

@aspirisen what do you think about it?

packages/slate-editor/src/modules/editor-v4-floating-add-menu/FloatingAddMenu.tsx
aspirisen: Maybe put it here packages/slate-editor/src/lib ?

packages/slate-editor/src/modules/editor-v4-floating-add-menu/FloatingAddMenu.tsx
aspirisen: hm, personally i am not against current behaviour, i would leave it, at least for now

aspirisen:
Shaun The Sheep Movie Ok GIF

aspirisen (approved):

packages/slate-editor/src/modules/editor-v4-floating-add-menu/FloatingAddMenu.tsx
OskarGasiorowski: @aspirisen hmm, let's leave as it is, if we would need in another place we then move it :)

PR #157, opened 2022-04-28
[MT-4930] Fix - Visible caret in non-editable elements in Safari

OskarGasiorowski:
This is what it looked like:
https://user-images.githubusercontent.com/41072665/165730953-48efe467-8469-4d65-8ced-c6193a7bdb13.mp4


packages/slate-editor/src/components/EditorBlock/EditorBlock.module.scss
```diff
@@ -4,6 +4,7 @@
 .outer {
     @include editor-v4-block($void: false);
     position: relative;
+    caret-color: transparent;
```
e1himself, line 7: This makes the cursor invisible in Image caption fields.

e1himself (changes requested):

packages/slate-editor/src/components/EditorBlock/EditorBlock.module.scss
OskarGasiorowski: Nice, you found this, thanks. I will fix it :)

e1himself: Fixed with 201aa6a7f1106dc21977d6e1c64cad7b37bf8d17

e1himself (approved):

PR #161, opened 2022-05-04
Add empty state for story bookmark

aspirisen:


packages/slate-editor/src/modules/editor-v4-story-bookmark/components/StoryBookmarkElement/StoryBookmarkElement.tsx
```diff
@@ -52,51 +52,77 @@ export function StoryBookmarkElement({ attributes, children, element, params }:
...
-                        withNewTabOption={params.withNewTabOption}
-                        onUpdate={(attrs) => updateImage(editor, attrs)}
-                        onRemove={remove}
-                    />
-                )
+            overlay={hasStory ? 'always' : false}
+            renderMenu={
+                hasStory
+                    ? ({ onClose }) =>
+                          story && (
```
e1himself, line 65: I can see that there's another condition for menu rendering within the render function: `story && (`.

Can it not be combined with the `hasStory` condition to have less indents?

packages/slate-editor/src/components/Button/Button.module.scss
```diff
@@ -50,7 +50,7 @@
         opacity: 1;
     }
 
-    &--disabled,
+    .--disabled,
```
e1himself, line 53: Hm... Should it not be `&.--disabled`? :thinking:

packages/slate-editor/src/components/Button/Button.module.scss
```diff
@@ -85,7 +85,29 @@
...
+            opacity: 0.4;
+        }
+    }
+
+    &--secondary {
+        background: $grey-200;
+        color: #4b5563;
+        padding: $spacing-1 - $border-size;
+
+        &:not(&--disabled, &:disabled) {
```
e1himself, line 99: `&--disabled`

I think this one is not aligned with other `--disabled` classes.

packages/slate-editor/src/components/ElementPlaceholder/ElementPlaceholder.stories.tsx
```diff
@@ -0,0 +1,31 @@
...
+                        <ChickenNoSignalIllustration />
+                    </Illustration>
+                    <VStack spacing="1">
+                        <Title>The selected Prezly Story is no longer available</Title>
+                        <Subtitle>
+                            Click to select another a bookmark of one of your stories
+                        </Subtitle>
+                    </VStack>
+                </VStack>
+            </Container>
```
e1himself, line 28: As always, I am not a fan of constructor components. I'd better see this a black-box component with input properties. There's no need to make it more complex than that at this stage. KISS.

packages/slate-editor/src/modules/editor-v4-story-bookmark/components/StoryBookmarkElement/StoryBookmarkElement.tsx
```diff
@@ -52,51 +52,77 @@ export function StoryBookmarkElement({ attributes, children, element, params }:
...
+                                isSelected={isSelected}
+                                element={element}
+                                story={story}
+                            />
+                        </div>
+                    );
+                }
+
+                return (
+                    <div>
```
e1himself, line 112: I wonder what these divs are needed for?

packages/slate-editor/src/lib/stripTags.tsx
```diff
@@ -0,0 +1,9 @@
+import striptags from 'striptags';
+
+export function stripTags(text?: string) {
+    if (!text) {
+        return null;
```
OskarGasiorowski, line 5: I am not a fan of returning `null`. Could we return just an empty string here?

OskarGasiorowski (approved):
Small thing could be changed in my opinion, but it's still looks good for me :)

packages/slate-editor/src/components/Button/Button.module.scss
e1himself: @aspirisen what about this?

packages/slate-editor/src/components/ElementPlaceholder/ElementPlaceholder.stories.tsx
e1himself: I think we're talking about complexity on different abstraction levels. Let me try to convince you by explaining my perspective on this :)

One of the base principles of software development is to hide complexity by abstracting it into something simple enough for people to use. This applies to UIs, APIs, and components as well.

By exposing all the internals of this placeholder component, you force the consumer to: 
- learn which building blocks the component has
- find an example of how they can be composed
- import 4 components: `Container`, `Title`, `Subtitle`, and `VStack`
- composed them in the correct order, making sure they do not produce an inconsistent look & feel

This is the complexity I'm talking about. 

In contrast, if it was a black-box component with properties exposing its inputs, it becomes trivial: you just provide all details and you get a consistent result with no knowledge of the internals. 

There are, of course, cases when it's impossible to avoid exposing this complexity (like UI libraries, or frameworks). But for me, the rule of thumb is to always avoid that unless it becomes necessary. In this particular case, there are no signs yet that someone will need to build a different combination of `Container`, `Title`, `Subtitle` and `VStack` structure. Thus I'd better go with a simpler-to-use implementation. Encapsulated black-box it is.

Thoughts?

aspirisen: I see your point, but looks like we have different view on such components. I treat them as dumb renders like ui libraries. For examples here in story we have `Click to select another a bookmark of one of your stories` but for specific reason in the editor itself we do not have this line, so we would need to add render condition in the component what imo reads harder that JSX. Also according to designs we will have DND in such components and we will need add special components to handle such behavior and therefore make this component to accept more and more props. 

We are adding ui to the page even when we are not working on ui library, but face the same issues. I would like to follow common approaches on building ui components even if when we are not working on ui-kit because in general we are solving the same issues. 

Your points are true, but in my opinion they should be applied to components with business or complex logic that depends on other sources.

In my mind there should be components package that operates UI, and other parts that uses them to build UI and even do not create their own css, all css should be in the ui components. If there is a collision it means that there is a mistake in designs that they added smth extraordinary and it could be solved by our already created components, or there are really new changes and we need update our components. There should not be a case that we create special UI behavior like hover color, button background etc. for a page just because we have it in designs.

It looks like we have controversial opinion on the same base things and they can't be resolve in scope of this PR. I will change the code here to have one component, but would like to discuss approach of building ui components in future.

aspirisen:
@e1himself could you take a look again

packages/slate-editor/src/components/ElementPlaceholder/Illustration.tsx
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+import type { PropsWithChildren } from 'react';
+
+import styles from './ElementPlaceholder.module.scss';
+
+interface Props {}
+
+export function Illustration(props: PropsWithChildren<Props>) {
+    return <div className={styles.illustration}>{props.children}</div>;
+}
```
e1himself, line 10: As of my taste, these one-liner components do not deserve a separate file. We can just inline them into the `EmailPlaceholder.tsx` file. 

```tsx
function Illustration(props: { children: ReactNode }) {
    return <div className={styles.illustration}>{props.children}</div>;
}
```

e1himself (approved):
Looks good! Thanks Dmitry! :+1:

PR #163, opened 2022-05-10
[MT-4929] Fix - Doubling divider behaviour on copy/paste

OskarGasiorowski:
Some clarification. This is what we get when we paste the `divider` to Safari (or to another browser copying from Safari). Slate treats this node as two different ones. It recognizes `data-slate-node` as one and `hr` as another.
```
<div data-slate-node="element" data-slate-void="true" class="_outer_1o71p_61 _void_1o71p_73 _extendedHitArea_1o71p_65" data-slate-type="divider" data-slate-value="{&quot;children&quot;:[{&quot;text&quot;:&quot;&quot;}],&quot;type&quot;:&quot;divider&quot;}" data-element-layout="contained">
  <div class="_card_1o71p_69 _alignCenter_1o71p_98" contenteditable="false">
    <hr class="_divider_9b6vx_31">
  </div>
  <div data-slate-spacer="true" style="height: 0px; color: transparent; outline: none; position: absolute;"><span data-slate-node="text"><span data-slate-leaf="true"><span data-slate-zero-width="z" data-slate-length="0">&#xFEFF;</span></span></span></div>
</div>
```
In that place we didn't know if `hr` is from Slate or just from some HTML code:
```
[DIVIDER_NODE_TYPE]: createDeserializeElement(parseSerializedElement),
HR: createDivider,
```
So I've added a flag to mark elements that come from slate and can be rendered by the parent slate node.


e1himself (approved):
Not too elegant, but does the job. Well done Oskar :+1:

PR #172, opened 2022-05-13
[MT-4965] Fix - Image menu overlapping with caption

e1himself:
Improve Image menu behavior vs caption-text selection

- Do not open the menu every time image caption text is focused
- Automatically close the menu when caption text is clicked

### DIFF
--- a/packages/slate-editor/src/components/EditorBlock/EditorBlock.tsx
+++ b/packages/slate-editor/src/components/EditorBlock/EditorBlock.tsx
@@ -98,12 +98,6 @@ export const EditorBlock = forwardRef<HTMLDivElement, Props>(function (
         [editor, element],
     );
 
-    const handleNonVoidChildrenClick = useCallback(
-        function () {
-            setMenuOpen(false);
-        },
-        [],
-    );
 
     useEffect(
         function () {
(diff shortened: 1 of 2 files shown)


packages/slate-editor/src/components/EditorBlock/EditorBlock.tsx
```diff
@@ -82,24 +82,35 @@ export const EditorBlock = forwardRef<HTMLDivElement, Props>(function (
...
     const [container, setContainer] = useState<HTMLDivElement | null>(null);
-    const openMenu = useCallback(() => setMenuOpen(true), [setMenuOpen]);
-    const closeMenu = useCallback(() => setMenuOpen(false), [setMenuOpen]);
+    const closeMenu = useCallback(() => setMenuOpen(false), []);
 
-    const handleClick = useCallback(
-        function () {
-            openMenu();
+    const handleBlockClick = useCallback(
+        function (event) {
```
aspirisen, line 88: I think it is better to add type annotations because useCallback implicitly makes arguments to be any

aspirisen (approved):

packages/slate-editor/src/components/EditorBlock/EditorBlock.tsx
e1himself: Fixed. Thanks!

PR #174, opened 2022-05-13
[MT-4966] Fix scroll-into-view behavior for tall galleries

e1himself:
- Fix scroll-into-view logic for tall blocks
- Add an extra padding to the calculations, so there is some space between the block edge and the viewport edge

Before:
Peek 2022-05-13 16-50

After:

https://user-images.githubusercontent.com/370680/168305763-177d189d-c578-4a43-b4de-6cf25e22d228.mp4


e1himself:
Closing in favor of #149

PR #182, opened 2022-05-19
Refactor - Rich Formatting dropdown

e1himself:
- Disconnect Rich Formatting dropdown options rendering from the specific Elements components (less styles to override)
- Dramatically simplify Dropdown options rendering logic
- Change how Dropdown symbols are exported (it's a TS namespace now)
- Convert all related stylesheets to SCSS modules
- No changes in visual presentations are expected


packages/slate-editor/src/components/Menu/Dropdown.module.scss
```diff
@@ -132,54 +132,4 @@
...
-            padding: 0;
-            background: unset;
-        }
-
-        &--blockquote > p {
-            padding-left: $spacing-half;
-            font-style: italic;
-        }
-
-        &--blockquote > p::before {
```
e1himself, line 177: This part was a spaghetti blob mixed with the code located on a different abstraction level. Fixed by moving the styles around to where they belong.

.eslintrc.cjs
```diff
@@ -62,6 +62,7 @@ module.exports = {
         "@typescript-eslint/consistent-type-imports": "error",
         "@typescript-eslint/no-empty-interface": "off",
         "@typescript-eslint/no-explicit-any": "off",
+        "@typescript-eslint/no-namespace": "off",
```
aspirisen, line 65: 👍

packages/slate-editor/src/components/Menu/Dropdown.tsx
```diff
@@ -1,31 +1,35 @@
...
+        hidden?: boolean;
+        label: string;
+        value: Value;
+    }
 
-export interface Props<Value extends string> extends Omit<DropdownProps, 'onChange'> {
-    onChange: (value: Value) => void;
-    options: Option<Value>[];
-    value?: Value;
+    export interface Props<Value extends string> extends Omit<DropdownProps, 'onChange'> {
```
aspirisen, line 16: Is it new approach that we should follow from now?

aspirisen (approved):

aspirisen:
asd

packages/slate-editor/src/components/Menu/Dropdown.tsx
e1himself: Yes. No. I don't know. Let's talk :) 

I wanted to group the dropdown with its related types, and tried this approach. It's nicer than prefixing all the types with `Dropdown`. Nicer than ES namespaces (though not native). 

What do you think? Yay or neigh?

aspirisen: I would like to give it a try, for now i do not see disadvantages 😊

packages/slate-editor/src/components/Menu/Dropdown.module.scss
```diff
@@ -52,9 +52,23 @@
...
+                content: "";
+                display: block;
+                position: absolute;
+                left: $spacing-base;
+                top: $spacing-base;
+                bottom: $spacing-base;
+                right: $spacing-base;
+                border-radius: $border-radius-base;
+                background-color: $toolbar-dark-theme-hover-bg;
+            }
```
e1himself, line 68: It turned out that the hover state was missing here, and instead was applied from the main app stylesheet :grimacing:

PR #183, opened 2022-05-20
[MT-4981] Fix File Attachment double border

e1himself:


packages/slate-editor/src/extensions/file-attachment/components/FileAttachment.module.scss
```diff
@@ -2,30 +2,12 @@
...
-        @include editor-v4-block-active;
-    }
-
-    &--unstyled {
-        box-shadow: none;
-    }
-}
-
-.content {
-    @include editor-v4-void-element();
```
e1himself, line 22: The problem was that the `.file-attachment-element` here was actually styled as a top-level frame. Which was unnecessary, as we've wrapped it with `EditorBlock` already.

aspirisen (approved):

PR #195, opened 2022-05-27
[MT-4990] Fix full-width block layout & selection styling

e1himself:


packages/slate-editor/src/components/EditorBlock/EditorBlock.module.scss
```diff
@@ -81,6 +81,23 @@
...
+            display: block;
+            position: absolute;
+            left: 0;
+            top: 0;
+            right: 0;
+            bottom: 0;
+            pointer-events: none;
+            outline: solid 6px $editor-block-focus-color;
+            outline-offset: -6px;
+        }
```
e1himself, line 99: This is a trick to have inset outline that will not be covered with children elements contents.

aspirisen (approved):

PR #196, opened 2022-05-27
[MT-4991] Fix - iframely embed selection

e1himself:


packages/slate-editor/src/components/EditorBlock/EditorBlock.module.scss
```diff
@@ -57,7 +57,9 @@
 }
 
 .Content {
+    position: relative;
     overflow: hidden;
+    z-index: $editor-block-overlay-z-index - 1;
```
e1himself, line 62: Apparently, Iframely-injected website card markup has `z-index: 20`. Explicitly setting a lower `z-index` on the `.Content` root will make sure that any `z-index` set within it will never take priority over the `.Overlay` z-index.

aspirisen (approved):

PR #202, opened 2022-06-03
[MT-5017] Fix - Use image original src as a return value from the `useImage` hook

e1himself:
See comments in MT-5017 for details.

- Change `useImage` return object to provide `loaded` flag, instead of the blob URL
- Rename `useImage` internal functions for clarity

### DIFF
--- a/packages/slate-editor/src/components/LoadingPlaceholder/ImageWithLoadingPlaceholder.tsx
+++ b/packages/slate-editor/src/components/LoadingPlaceholder/ImageWithLoadingPlaceholder.tsx
@@ -48,7 +48,7 @@ export const ImageWithLoadingPlaceholder = forwardRef<HTMLDivElement, Props>((pr
     } = props;
     const aspectRatio = imageHeight ? imageWidth / imageHeight : undefined;
     const callbacks = useLatest({ onStartLoading, onLoad });
-    const { loading, progress, url } = useImage(src);
+    const { loading, progress, loaded: isLoaded } = useImage(src);
 
     useEffect(
         function () {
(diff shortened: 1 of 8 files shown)


packages/slate-editor/src/lib/hooks/useImage.ts
```diff
@@ -12,7 +12,7 @@ interface State {
 }
 
 export function useImage(src: string): State {
-    const getPromise = useCallback(() => createImageProgressPromise(src), [src]);
+    const getPromise = useCallback(() => fetchImageWithProgress(src).then(() => src), [src]);
```
e1himself, line 15: This is the actual fix. I've added `.then(() => src)`.

packages/slate-editor/src/components/LoadingPlaceholder/ImageWithLoadingPlaceholder.tsx
```diff
@@ -73,12 +73,12 @@ export const ImageWithLoadingPlaceholder = forwardRef<HTMLDivElement, Props>((pr
...
             style={{
                 aspectRatio: aspectRatio ? String(aspectRatio) : undefined,
-                backgroundImage: url ? `url("${url}")` : undefined,
+                backgroundImage: isLoaded ? `url("${src}")` : undefined,
                 ...attributes.style,
             }}
             ref={ref}
         >
-            {url && <img className={styles.Image} src={url} alt={alt} />}
+            {isLoaded && <img className={styles.Image} src={src} alt={alt} />}
```
e1himself, line 81: This is the fix. Instead of using the "blob" `url` returned from `useImage`, we rely on the original `src` string. While `useImage` is still used for progress reporting.

PR #211, opened 2022-06-08
Refactor root type normalization

aspirisen:


e1himself (approved):
:+1: 

- A few minor comments below
- Is there a single test for this functionality? Would it be worth writing one?

e1himself (approved):

PR #223, opened 2022-06-22
[MT-5193] PoC - Entry point above the first block node

e1himself:
https://user-images.githubusercontent.com/370680/174981560-48e23575-4a18-4d25-b3ca-a8a31f2a579c.mp4


e1himself:
Closing in favor of an alternative solution: #265

PR #228, opened 2022-06-25
Add cell content formatting

aspirisen:


packages/slate-editor/src/modules/rich-formatting-menu/components/Toolbar.tsx
```diff
@@ -75,9 +77,11 @@ export function Toolbar({
     return (
         <>
             <Menu.ButtonGroup>
-                <Menu.Button active={isBold} onClick={onBold}>
-                    <Menu.Icon icon={FormatBold} />
-                </Menu.Button>
+                {withBoldFormat && (
```
aspirisen, line 80: This is not the best solution as you can copy-paste bold text in header cell, in this case we need to adopt normalization to remove bold mark

PR #235, opened 2022-07-11
[MT-5070] Fix - Cursor works incorrectly in table cell

e1himself:


packages/slate-editor/src/components/EditorBlock/EditorBlock.tsx
```diff
@@ -43,7 +43,12 @@ export interface Props extends Omit<RenderElementProps, 'attributes'>, SlateInte
     hasError?: boolean;
     layout?: `${Layout}`;
     overlay?: OverlayMode;
-    renderBlock: (props: { isSelected: boolean }) => ReactNode;
+    renderAboveFrame?: FunctionComponent<{ isSelected: boolean }> | ReactNode;
+    renderBelowFrame?: FunctionComponent<{ isSelected: boolean }> | ReactNode;
+    renderEditableFrame?: FunctionComponent<{ isSelected: boolean }>;
+    renderReadOnlyFrame?: FunctionComponent<{ isSelected: boolean }>;
+    // renderEditableFrame: ...
```
aspirisen, line 50: Do we need these comments?

packages/slate-editor/src/components/EditorBlock/EditorBlock.tsx
```diff
@@ -72,6 +80,16 @@ export const EditorBlock = forwardRef<HTMLDivElement, Props>(function (
     },
     ref,
 ) {
+    if (typeof children !== 'undefined') {
```
aspirisen, line 83: Why do we need runtime check, TS will throw an error if you pass some non-undefined children

packages/slate-editor/src/lib/hooks/useFunction.ts
```diff
@@ -6,7 +6,7 @@ import { useLatest } from './react-use';
  * Return a stable proxy-function
  * that will always call the latest version of the input function.
  */
-export function useFunction<F extends () => any>(func: F): F {
+export function useFunction<F extends (...params: any) => any>(func: F): F {
```
aspirisen, line 9: Could you use `...params: never[]`
because if you will forgive to set type it will be `any` without errors
image

packages/slate-editor/src/extensions/story-embed/types.ts
```diff
@@ -1,22 +1,15 @@
 import type { StoryEmbedNode } from '@prezly/slate-types';
-import type { ReactNode } from 'react';
+import type { FunctionComponent } from 'react';
```
aspirisen, line 2: nit: you can use `ComponentType` just in case of classic class components

e1himself: Yeah, I've tried that. See 5bc526dc3a6be5fb69aca53cffe9fa51c97678fd :)

aspirisen: hm, if it is not a component i would avoid using *Component type, just plain callback function

packages/slate-editor/src/components/EditorBlock/EditorBlock.tsx
e1himself: No, it won't.

This is because `children` is declared as `any` in `RenderElementProps` from `slate-react`. 

`any` passes all checks, including `never` and `undefined`.

aspirisen: why not just remove `children` from props type declaration?

e1himself: Worked! Thanks! :trophy: 

45293b6ca35046e8aaccf0dde6913d2ea5a114e9

packages/slate-editor/src/extensions/story-embed/types.ts
e1himself: Done in 54a46704fba2b7a4c4dde5323fee5284e9f7a3c6. Thanks! :+1:

aspirisen (approved):

PR #239, opened 2022-07-11
Add track events for table

aspirisen:


packages/slate-tables/src/nodes/TableNode.ts
```diff
@@ -77,6 +77,6 @@ export namespace TableNode {
         // and bold mark still present in cell content
         Editor.normalize(editor, { force: true });
 
-        return true;
+        return { newValue: !isAlreadyEnabled, headerType };
```
e1himself, line 80: > `{ newValue: !isAlreadyEnabled, headerType }`

I think the result of this generic function is too much tailed for the analytics tracking even payload. 
I'd rather make it return the new `header` property value and do the rest of the calculations where the tracking happens.

Here: 
```ts
return newHeader;
```

Where the tracking happens:

```ts
const header = TablesEditor.toggleTableHeader(
      editor,
      undefined,
      'first_row',
  );

  if (toggled) {
      EventsEditor.dispatchEvent(editor, 'table-toggle-header', {
          headerType: 'first_row',
          newValue: header.includes('first_row'),
      });
  }
```

PR #247, opened 2022-07-14
Fix coping from table cell

aspirisen:


packages/slate-editor/src/extensions/tables/TablesExtension.tsx
```diff
@@ -64,24 +65,26 @@ export function TablesExtension(): Extension {
...
-                            children: children as TableRowNode['children'] | undefined,
-                        }),
-                    createTableCellNode,
-                    isTableNode,
-                    isTableRowNode,
-                    isTableCellNode,
-                }),
+            return withTablesCopyPasteBehavior(
+                withTablesDeleteBehavior(
+                    withTables(editor, {
```
e1himself, line 70: I guess it's time to use `flow()` to unfold these nesting levels.

See https://github.com/prezly/slate/blob/82ef1062880eca584a647e7bd3ac2019005c73b0/packages/slate-editor/src/modules/editor/createEditor.ts#L29-L44

packages/slate-tables/src/core/withTablesCopyPasteBehavior.ts
```diff
@@ -0,0 +1,50 @@
...
+            }
+        }
+
+        return getFragment();
+    };
+
+    return editor;
+}
+
+function findNearestCell(editor: TablesEditor, at: Location) {
```
e1himself, line 42: `findParentCell` maybe? "Nearest" can mean a lot more than just looking up the tree.

e1himself (approved):
Looks great! 

I only have a couple of suggestions on how to make the code slightly more readable. Otherwise, good job! :+1:

PR #250, opened 2022-07-19
[MT-5131] Feature - [+] menu "Suggestions" group

e1himself:
Screenshot from 2022-07-19 19-41-03

### DIFF
--- a/packages/slate-editor/src/components/FancyScrollbars.tsx
+++ b/packages/slate-editor/src/components/FancyScrollbars.tsx
@@ -1,6 +1,5 @@
 import React, { Component } from 'react';
-import type { ScrollbarProps } from 'react-custom-scrollbars-2';
-import { Scrollbars } from 'react-custom-scrollbars-2';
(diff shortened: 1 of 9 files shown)


packages/slate-editor/src/components/FancyScrollbars.tsx
```diff
@@ -84,17 +83,24 @@ export class FancyScrollbars extends Component<Props> {
...
-        return getScrollParent(element.parentElement);
-    }
+        if (parent === null) return null;
+
+        if (parent.scrollHeight > parent.clientHeight && isScrollable(parent)) {
+            return parent;
+        }
+
+        current = parent;
+    } while (current);
```
e1himself, line 98: There was a bug here.

For me, it was selecting the menu item itself as its scroll parent (i.e. it was a bug in `getScrollParent()`). Checking `overflowY` and excluding the current element is the fix.

Additionally, I've rewritten the function to a loop, instead of a recursion (yeah, it's a micro-optimization).

aspirisen (approved):

PR #256, opened 2022-07-20
[MT-5175] Add normalization to fix broken tables

aspirisen:


packages/slate-tables/src/normalization/normalizeTableData.ts
```diff
@@ -0,0 +1,89 @@
+import { omit } from 'lodash-es';
+import type { Path } from 'slate';
+import { Node, Transforms } from 'slate';
+import { Editor } from 'slate';
+
+import { TablesEditor } from '../TablesEditor';
+
+export function normalizeTableData(editor: TablesEditor, path: Path) {
```
e1himself, line 8: If I'm not mistaken, this normalization splits merged cells (with `colspan` / `rowspan` attributes) into single 1x1 cells, right?

If so, I'd recommend renaming the function to make that obvious. `normalizeTableData` is too generic and can be used for any other normalization introduced in this PR.

How about `splitMergedCells`?

I also think that this function is too big and too complex to understand. I'd suggest splitting it into two: 

- `splitColSpanCells`
- `splitRowSpanCells`

```ts
export function splitColSpanCells(editor: TablesEditor, path: Path) {
    const node = Node.get(editor, path);

    if (editor.isTableCellNode(node) && node.colspan && node.colspan > 1) {
        const padCells = times(node.colspan - 1).map(() => TablesEditor.createTableCell(editor));

        Editor.withoutNormalizing(editor, () => {
            Transforms.unsetNodes<TableCellNode>(editor, 'colspan', { at: path });
            Transforms.insertNodes(editor, padCells, { at: Path.next(path) });
        });
        return true;
    }

    return false;
}

export function splitRowSpanCells(editor: TablesEditor, path: Path) {
    const node = Node.get(editor, path);

    if (editor.isTableNode(node)) {
        for (const [cell, cellPath] of TablesEditor.cells(editor, [node, path])) {
            if (cell.rowspan && cell.rowspan > 1) {
                const [rowIndex, cellIndex] = cellPath.slice(-2);
                const padCells = times(cell.rowspan - 1).map(() =>
                    TablesEditor.createTableCell(editor),
                );

                Editor.withoutNormalizing(editor, () => {
                    Transforms.unsetNodes<TableCellNode>(editor, 'rowspan', { at: cellPath });
                    for (const [offset, padCell] of padCells.entries()) {
                        // TODO: Might be worth checking if the insertion path exists.
                        Transforms.insertNodes(editor, padCell, {
                            at: [...path, rowIndex + 1 + offset, cellIndex],
                        });
                    }
                });

                return true;
            }
        }
    }

    return false;
}
```

packages/slate-editor/src/modules/nodes-hierarchy/tests/Tables.test.tsx
```diff
@@ -303,4 +303,245 @@ describe('nodes-hierarchy / Tables', () => {
 
         expect(editor.children).toEqual(expected.children);
     });
+
+    it('should remove empty row', () => {
```
e1himself, line 307: > Tables.test.tsx

Ideally, this test should live inside `slate-tables` package directory. As it's only testing the code inside `slate-tables` package.

e1himself (commented):
Looks functional :D :+1: 

I have a few comments on improving naming, code separation and code location as well. 
Please see below.

packages/slate-editor/src/modules/nodes-hierarchy/hierarchySchema.ts
```diff
@@ -94,6 +94,14 @@ export const hierarchySchema: NodesHierarchySchema = {
...
     [TABLE_NODE_TYPE]: [
         allowChildren(isTableRowNode, combineFixers([unwrapSameTypeChild, liftNodeNoSplit])),
+        (editor, path) =>
+            TablesEditor.isTablesEditor(editor) && TablesNormalization.removeEmptyRow(editor, path),
+        (editor, path) =>
+            TablesEditor.isTablesEditor(editor) &&
+            TablesNormalization.normalizeTableData(editor, path),
+        (editor, path) =>
+            TablesEditor.isTablesEditor(editor) &&
+            TablesNormalization.restoreAbsentCells(editor, path),
```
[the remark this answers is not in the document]
aspirisen: imo i would move all normalizations binding to the editor, and export normalizators from package. In this case we can see all in one place, can disable if we need, change order (put smth in the middle because order matters), override default normalization etc.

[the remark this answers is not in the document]
aspirisen: here for example new normalization can work only after wrong nodes will be unwrapped or lifted

packages/slate-tables/src/normalization/normalizeTableData.ts
aspirisen: i am not sure that it is possible to make two different functions as i have to restore cells in on loop

e1himself: > as i have to restore cells in one loop

Hm.. Why?

packages/slate-editor/src/modules/nodes-hierarchy/hierarchySchema.ts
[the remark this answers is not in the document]
e1himself: > here for example new normalization can work only after wrong nodes will be unwrapped or lifted

I read this as "*the hierarchy normalization should run before extension normalizations*" :)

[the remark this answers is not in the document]
aspirisen: But how can we control the order, if is just override editor.normalize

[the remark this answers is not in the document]
e1himself: > But how can we control the order, if is just override editor.normalize

I'd say we just put the hierarchy normalization last in this list:

```diff
packages/slate-editor/src/modules/editor/createEditor.ts
@@ -35,7 +35,6 @@
         withReact,
         withHistory,
         withNonEmptyValue,
-        withNodesHierarchy(hierarchySchema),
         withBreaksOnExpandedSelection,
         withBreaksOnVoidNodes,
         withInlineVoid(getExtensions),
@@ -47,5 +46,6 @@
         withSerialization(getExtensions),
         ...overrides,
         ...plugins,
+        withNodesHierarchy(hierarchySchema),
     ])(baseEditor);
 }
```

packages/slate-editor/src/modules/nodes-hierarchy/tests/Tables.test.tsx
aspirisen: There is a set-up for jsx and types in this package, i would leave this tests here for now and will think how to share elements bindings among pacakges

e1himself: Alright. Let's address this later.

> how to share elements bindings among packages

For the record, I'd better not overcomplicate the setup with shared bindings. A local partially duplicated setup inside `slate-tables` can be simpler if nothing is shared.

packages/slate-tables/src/withNormalization.ts
```diff
@@ -0,0 +1,29 @@
+import * as TablesNormalization from './normalization';
+import type { TablesEditor } from './TablesEditor';
+
+const normalizers = [
+    TablesNormalization.removeEmptyRows,
+    TablesNormalization.splitRowSpanCells,
+    TablesNormalization.splitColSpanCells,
+    TablesNormalization.insertMissingCells,
```
e1himself, line 8: nit: there's not much value in namespacing the normalization module functions.

e1himself (approved):

PR #257, opened 2022-07-21
[MT-5180] Fix - UX for copying/cutting/pasting tables

e1himself:
- Re-enable forgotten extension of `PasteSlateContent` (see #210) (improves pasting for tables)
- Changes to copy/cut/paste behaviour:
  - ~Copy the full table if multiple cells are selected~ (see [the discussion](https://github.com/prezly/slate/pull/257#discussion_r927127028))
  - Copy/cut the full table if selection is collapsed and the cursor is inside table cell
- Unbind `TablesExtension` from `ParagraphsExtension` -- it's no longer dependent on it directly

### DIFF
--- a/packages/slate-editor/src/extensions/paste-slate-content/PasteSlateContentExtension.ts
+++ b/packages/slate-editor/src/extensions/paste-slate-content/PasteSlateContentExtension.ts
@@ -2,7 +2,7 @@ import type { Extension } from '@prezly/slate-commons';
 
 import { withSlatePasting } from './lib';
 
-export const EXTENSION_ID = PasteSlateContentExtension.name;
+export const EXTENSION_ID = 'PasteSlateContentExtension';
 
 export function PasteSlateContentExtension(): Extension {
     return {
--- a/packages/slate-editor/src/extensions/paste-slate-content/PasteSlateContentExtension.ts
+++ b/packages/slate-editor/src/extensions/paste-slate-content/PasteSlateContentExtension.ts
@@ -2,7 +2,7 @@ import type { Extension } from '@prezly/slate-commons';
 
 import { withSlatePasting } from './lib';
 
-export const EXTENSION_ID = 'PasteSlateContentExtension';
+export const EXTENSION_ID = PasteSlateContentExtension.name;
 
 export function PasteSlateContentExtension(): Extension {
     return {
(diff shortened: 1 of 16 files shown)


aspirisen:
Why do we need to copy whole table if we select some cells?

packages/slate-editor/src/extensions/tables/onKeyDown.ts
```diff
@@ -0,0 +1,52 @@
...
+        if (!entry) {
+            return;
+        }
+        const [, path] = entry;
+
+        Editor.withoutNormalizing(editor, () => {
+            Transforms.select(editor, path); // Select the table
+            const ref = Editor.pathRef(editor, path);
+
+            if (document.execCommand('cut')) {
```
aspirisen, line 41: if you select several cells and then cut them the cells will be removed as well and table can be shifted. I think we need just copy selected content and then remove selected nodes, without cells

e1himself: > Why do we need to copy whole table if we select some cells?

Hm... For me, it never does what I expect (in Firefox).
However, I've tried it in Chrome and it's much better. So probably you're right that it's better to revert it.

> I think we need to just copy selected content and then remove selected nodes, without cells

Oh, good idea. I like it :+1: Will do.

e1himself: > if you select several cells and then cut them the cells will be removed as well and table can be shifted

This particular code here is only running when the selection is collapsed. 

- If the cursor is inside a cell, the selection is collapsed, and "copy" hotkey is pressed, we copy the whole table
- If the cursor is inside a cell, the selection is collapsed, and "cut" hotkey is pressed, we cut the whole table

I believe this consistency does make sense.

> Copy the full table if multiple cells are selected

This part was implemented in `withTablesCopyPasteBehaviour` and is reverted by eb1bfd91c0d4b1011f7eea582ccb7c217a81aafe.

Thanks!

packages/slate-editor/src/extensions/paste-slate-content/PasteSlateContentExtension.ts
```diff
@@ -2,7 +2,7 @@ import type { Extension } from '@prezly/slate-commons';
 
 import { withSlatePasting } from './lib';
 
-export const EXTENSION_ID = 'PasteSlateContentExtension';
+export const EXTENSION_ID = PasteSlateContentExtension.name;
```
aspirisen, line 5: q: i am not sure, but maybe uglifier can change the name of function and the name can be random

e1himself: Good point. Will fix. Thanks :+1:

aspirisen (approved):

PR #258, opened 2022-07-21
[MT-5197] Fix - Correct checking of edge line in Table cells

aspirisen:

### DIFF
--- a/packages/slate-commons/src/commands/isCursorOnEdgeOfContainer.ts
+++ b/packages/slate-commons/src/commands/isCursorOnEdgeOfContainer.ts
@@ -4,14 +4,26 @@ import { ReactEditor } from 'slate-react';
 
 export type ContainerEdge = 'top' | 'bottom';
 
+export function isCursorOnFirstLine(editor: ReactEditor, container: Point, cursor: Point): boolean {
+    return isCursorOnEdgeOfContainer(editor, container, cursor, 'top');
+}
+
+export function isCursorOnLastLine(editor: ReactEditor, container: Point, cursor: Point): boolean {
+    return isCursorOnEdgeOfContainer(editor, container, cursor, 'bottom');
+}
+
 export function isCursorOnEdgeOfContainer(
     editor: ReactEditor,
     container: Point,
-    child: Point,
+    cursor: Point,
     edge: ContainerEdge,
 ) {
     const a = getPointRect(editor, container);
-    const b = getPointRect(editor, child);
+    const b = getPointRect(editor, cursor);
+
+    if (!a || !b) {
+        return false;
+    }
 
     switch (edge) {
         case 'top':
@@ -23,9 +35,16 @@ export function isCursorOnEdgeOfContainer(
 
 function getPointRect(editor: ReactEditor, point: Point) {
     const range = Editor.range(editor, { ...point, offset: Math.max(point.offset, 0) });
-    return getRangeRect(editor, range);
+    try {
+        return getRangeRect(editor, range);
+    } catch {
+        return undefined;
+    }
 }
 
+/**
+ * @throws error when `ReactEditor.toDOMRange()` cannot match range to a DOM node
+ */
 function getRangeRect(editor: ReactEditor, range: Range) {
     const domRange = ReactEditor.toDOMRange(editor, range);
     const rects = domRange.getClientRects();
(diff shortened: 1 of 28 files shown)


packages/slate-commons/src/commands/isCursorOnEdgeOfContainer.ts
```diff
@@ -0,0 +1,40 @@
...
+    const domRange = ReactEditor.toDOMRange(editor, range);
+    const rects = domRange.getClientRects();
+
+    // if the cursor will be in the beginning of next line there will be two rects:
+    // first the ending of previous line and the second one is the beginning of current line
+    if (rects.length) {
+        return Array.from(rects);
+    }
+
+    return [domRange.getBoundingClientRect()];
```
e1himself, line 39: I would love to avoid checking DOM client rects here, but apparently there's no way around this :disappointed:

packages/slate-commons/src/commands/isCursorOnEdgeOfContainer.ts
```diff
@@ -0,0 +1,40 @@
...
+    }
+}
+
+function getPointRect(editor: ReactEditor, point: Point) {
+    const range = Editor.range(editor, { ...point, offset: Math.max(point.offset, 0) });
+    return getRangeRect(editor, range);
+}
+
+function getRangeRect(editor: ReactEditor, range: Range) {
+    const domRange = ReactEditor.toDOMRange(editor, range);
```
e1himself, line 30: This can throw an error if ReactEditor cannot match range to DOM nodes. 

See #260

packages/slate-commons/src/commands/isCursorOnEdgeOfContainer.ts
```diff
@@ -0,0 +1,40 @@
+import { Editor } from 'slate';
+import type { Range, Point } from 'slate';
+import { ReactEditor } from 'slate-react';
+
+export type ContainerEdge = 'top' | 'bottom';
+
+export function isCursorOnEdgeOfContainer(
+    editor: ReactEditor,
+    container: Point,
+    child: Point,
```
e1himself, line 10: This argument should be renamed to `cursor`.

See #260

packages/slate-commons/src/commands/isCursorOnEdgeOfContainer.ts
```diff
@@ -0,0 +1,40 @@
+import { Editor } from 'slate';
+import type { Range, Point } from 'slate';
+import { ReactEditor } from 'slate-react';
+
+export type ContainerEdge = 'top' | 'bottom';
+
+export function isCursorOnEdgeOfContainer(
```
e1himself, line 7: I'd like to have shortcuts introduced (for better code readability):

- `isCursorOnFirstLine`
- `isCursorOnLastLine`

packages/slate-tables/src/core/onKeyDown.ts
```diff
@@ -47,27 +47,32 @@ function onUpPress(editor: TablesEditor): Point | undefined {
 
     const { activeCell, matrix } = traverse;
 
-    if (activeCell.row.isFirst) {
-        return Editor.before(editor, matrix.path, { unit: 'block' });
-    }
+    const [cellStart] = Editor.edges(editor, activeCell.path);
```
e1himself, line 50: It doesn't make sense to use `Editor.edges()` and only keep a part of the result. 

Just use `Editor.start()` directly.

From `slate` sources:

```ts
  /**
   * Get the start and end points of a location.
   */
  edges(editor, at) {
    return [Editor.start(editor, at), Editor.end(editor, at)];
  },
```

packages/slate-tables/src/core/onKeyDown.ts
```diff
@@ -86,27 +91,32 @@ function onDownPress(editor: TablesEditor): Point | undefined {
...
-    if (cellBelow) {
-        return EditorCommands.findLeafPoint(
-            editor,
-            {
-                path: cellBelow.path,
-                offset: editor.selection.anchor.offset,
-            },
-            'highest',
-        );
+    const [, cellEnd] = Editor.edges(editor, activeCell.path);
```
e1himself, line 94: Same as above. It doesn't make sense to use `Editor.edges()` and only keep a part of the result.

Use `Editor.end()` directly.

e1himself (approved):
All good. 

I have a couple of minor comments. Made a suggestion PR fixing those: #260. Please take a look.

e1himself (approved):

packages/slate-commons/src/commands/isCursorOnEdgeOfContainer.ts
aspirisen: me either 🥲

PR #266, opened 2022-07-26
Fix normalization logic

aspirisen:


packages/slate-editor/src/modules/nodes-hierarchy/normilizers/allowChildren.ts
```diff
@@ -10,10 +10,6 @@ export function allowChildren(
         const node = Node.get(editor, path);
 
         if ('children' in node) {
-            if (node.children.length === 0) {
-                return true;
```
e1himself, line 14: :zipper_mouth_face:

packages/slate-editor/src/modules/nodes-hierarchy/fixers/convertToParagraph.ts
```diff
@@ -12,6 +17,10 @@ export function convertToParagraph(editor: Editor, [node, path]: NodeEntry) {
         children.push(node);
     }
 
+    if (children.length === 0) {
+        children.push({ text: '' });
+    }
```
e1himself, line 22: I don't think this will work. 

IIRC, `Transforms.setNodes()` only sets node attributes and completely ignores `children`.

e1himself: https://github.com/ianstormtaylor/slate/blob/fe13a8f9e750569342ee004951e34233ab6614bf/packages/slate/src/transforms/node.ts#L658-L674

e1himself:
> Fix normalization logic

Can you please also describe what exactly was wrong?

aspirisen:
Slate has default normalization which inserts text if element has no children, here was a bug that this function was not called

packages/slate-editor/src/modules/nodes-hierarchy/normilizers/removeWhenNoChildren.ts
```diff
@@ -0,0 +1,16 @@
...
+        const node = Node.get(editor, path);
+
+        if ('children' in node && node.children.length === 0) {
+            Transforms.removeNodes(editor, { at: path, match: (n) => n === node });
+            return true;
+        }
+
+        return false;
+    };
+}
```
e1himself, line 16: Nitpicking.

Do you notice how `mustHaveChildren()` and `removeWhenNoChildren()` are almost the same?

I think they can be one *rule* with different fixes:

```ts
requireAtLeastOneChild(insertChild(createParagraph))
requireAtLeastOneChild(removeNode)
```

packages/slate-editor/src/modules/nodes-hierarchy/fixers/convertToParagraph.ts
e1himself: #267

packages/slate-editor/src/modules/nodes-hierarchy/fixers/insertTableRow.ts
```diff
@@ -0,0 +1,23 @@
...
+import { createTableCellNode, createTableRowNode } from '#extensions/tables';
+
+export function insertTableRow(editor: Editor, [node, path]: NodeEntry) {
+    Transforms.insertNodes(
+        editor,
+        [
+            createTableRowNode({
+                children: [createTableCellNode({ children: [createParagraph()] })],
+            }),
+        ],
```
e1himself, line 15: This is duplicating already existing code to construct rows/cells: `TableRowNode.createTableRow()`. 

I'd better expose `createTableRow` on the `TablesEditor` and use it here.

```suggestion
        TablesEditor.createTableRow(),
```

```diff
packages/slate-tables/src/nodes/TableRowNode.ts
===================================================================
@@ -20,7 +20,7 @@
             children:
                 typeof children === 'number'
                     ? Array.from(Array(children)).map(() => TableCellNode.createTableCell(editor))
-                    : children ?? [],
+                    : children ?? [TableCellNode.createTableCell(editor)],
         });
     }
```

packages/slate-editor/src/modules/nodes-hierarchy/normilizers/removeWhenNoChildren.ts
aspirisen: Yes, i thought about that, however this condition `requireAtLeastOneChild(removeNode)` makes me confused as it tells that it requires a child inside and at the same time removes whole node
I was thinking about conditions in naming
```
whenNoChild(insertChild(createParagraph))
whenNoChild(removeNode)
```
but as i remember we are not going this way

e1himself: > makes me confused as it tells that it requires a child inside and at the same time removes whole node

IDK, I see no problem. :man_shrugging:

packages/slate-editor/src/modules/nodes-hierarchy/fixers/insertTableRow.ts
e1himself: #269

PR #269, opened 2022-07-26
[MT-5209] Use row creation from table package

aspirisen:
See https://github.com/prezly/slate/pull/266#discussion_r930061068


packages/slate-editor/src/modules/nodes-hierarchy/fixers/insertTableRow.ts
```diff
@@ -1,23 +1,17 @@
...
-    Transforms.insertNodes(
-        editor,
-        [
-            createTableRowNode({
-                children: [createTableCellNode({ children: [createParagraph()] })],
-            }),
-        ],
-        {
+    if (TablesEditor.isTablesEditor(editor)) {
+        Transforms.insertNodes(editor, [TablesEditor.createTableRow(editor, { children: 3 })], {
```
e1himself, line 8: nit: it also accepts a single node. So wrapping into array is optional.

```ts
  insertNodes: <T extends Node>(
    editor: Editor,
    nodes: Node | Node[],
    options?: {
      at?: Location
      match?: NodeMatch<T>
      mode?: RangeMode
      hanging?: boolean
      select?: boolean
      voids?: boolean
    }
  ) => void
```

e1himself, line 8: Should we maybe just remove the table without any rows?

This will be consistent with what we do with empty rows in `slate-tables`: `removeEmptyRows()`.

packages/slate-tables/src/nodes/TableRowNode.ts
```diff
@@ -1,11 +1,11 @@
-import type { BaseElement, Location } from 'slate';
+import type { Element, Location } from 'slate';
 import { Transforms } from 'slate';
 
 import type { TablesEditor } from '../TablesEditor';
 
 import { TableCellNode } from './TableCellNode';
 
-export interface TableRowNode extends BaseElement {
+export interface TableRowNode extends Element {
```
e1himself, line 8: why? :)

e1himself (approved):

packages/slate-tables/src/nodes/TableRowNode.ts
aspirisen: Because we extent `Element` with `type` field

PR #276, opened 2022-07-27
[MT-5219] Fix - Editor inserting empty block before cropped image

e1himself:
The problem was here:

https://github.com/prezly/slate/blob/550124c7101cf67446cbad2559b799a86428bc7c/packages/slate-editor/src/modules/editor/lib/images/createImageEditHandler.ts#L41-L44

It should have been inserting a new paragraph at the top-most `Path` location, but `Selection.highest()` returned a `Range` for `editor.selection`. So it was inserting a paragraph inside another block.

This glitch was compensated by the previous normalization implementation. But not anymore, after we've refactored the normalization logic.

### DIFF
--- a/packages/slate-editor/src/extensions/loader/transforms/replaceLoader.ts
+++ b/packages/slate-editor/src/extensions/loader/transforms/replaceLoader.ts
@@ -1,5 +1,5 @@
 import { EditorCommands } from '@prezly/slate-commons';
-import type { Editor, Element } from 'slate';
+import { type Element, Editor } from 'slate';
 import { Transforms } from 'slate';
 
 import { findLoaderPath, isLoaderElement } from '../lib';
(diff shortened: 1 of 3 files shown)


packages/slate-editor/src/extensions/loader/transforms/replaceLoader.ts
```diff
@@ -12,15 +12,17 @@ export function replaceLoader(editor: Editor, loader: LoaderNode, element: Eleme
...
-        at: loaderPath,
-        mode: 'highest',
-    });
+        Transforms.insertNodes(editor, element, {
+            at: loaderPath,
+            mode: 'highest',
+        });
 
-    if (wasSelected) Transforms.select(editor, loaderPath);
+        if (wasSelected) Transforms.select(editor, loaderPath);
```
aspirisen, line 26: maybe at the end we still need to call normalization manually?

aspirisen (approved):

packages/slate-editor/src/extensions/loader/transforms/replaceLoader.ts
e1himself: > maybe at the end we still need to call normalization manually?

Why? 

https://docs.slatejs.org/api/nodes/editor#editor.withoutnormalizing-editor-editor-fn-greater-than-void-greater-than-void

> ```ts
> Editor.withoutNormalizing(editor: Editor, fn: () => void) => void
> ```
> 
> Call a function, deferring normalization until after it completes.

PR #279, opened 2022-07-28
[MT-5032] Fix - Editor crashing on pasting Table from a web-page

e1himself:
MT-5032


packages/slate-tables/src/normalization/insertMissingCells.ts
```diff
@@ -12,15 +12,9 @@ export function insertMissingCells(editor: TablesEditor, path: Path) {
         return false;
     }
 
-    const maxWidth = table.children.reduce((max, row) => {
-        const rowSize = calculateRowWidth(row);
```
e1himself, line 16: This line was assuming `table.children` are always `TableRowNode[]`. But sometimes it's not.

PR #286, opened 2022-08-19
[QRF24] Fix - Changing unordered list to paragraph

e1himself:
I recommend reviewing this with whitespaces ignored: https://github.com/prezly/slate/pull/286/files?diff=split&w=1

The core problem was that the `increaseDepth` and `decreaseDepth` functions have been producing slightly invalid document structure. But it wasn't visible, as the core `slate-list` extension normalizations were automatically fixing it. 

With the new hierarchy normalization added, it started to kick in while the `unwrapList` transformations were still running, producing additional side-effects. This made a new list node appear on every `unwrapList` iteration, causing the loop to run infinitely.

1. So I've made sure the tests for `increaseDepth` and `decreaseDepth` run without normalizations, so we can verify their behaviour independently of the normalization configuration.
2. Fixed their behavior to not insert unnecessary text nodes
  https://github.com/prezly/slate/blob/8a1ee6c8dba69c4d63ad5b64f7e3bb7031603c13/packages/slate-lists/src/lib/increaseListItemDepth.ts#L38
  https://github.com/prezly/slate/blob/8a1ee6c8dba69c4d63ad5b64f7e3bb7031603c13/packages/slate-lists/src/lib/decreaseListItemDepth.ts#L59
3. Additionally, I've added a hard protection against infinite loops here to at least make the problem visible in the future:
  https://github.com/prezly/slate/blob/8a1ee6c8dba69c4d63ad5b64f7e3bb7031603c13/packages/slate-lists/src/lib/unwrapList.ts#L14-L24

While learning the code, I've also refactored the cryptic `getUnreachableAncestors()` function:
- renamed it to `pickSubtreeRoots` (for clarity)
- added extensive documentation in its comments
- added a test for it
- move it into the `slate-list` package, as it wasn't used anywhere else


aspirisen (commented):
Good work 👍

PR #194, opened 2022-05-27
[MT-4987] Fix - Email editor WYSIWYG

e1himself:
Before:

screenshot-rock prezly com-2022 05 27-16_25_28

After:

image

PR #292, opened 2022-08-23
[DEV-7409] Fix - Pasting into Image caption

aspirisen:
As Image block is empty (despite that it has image) `insertFragment` removes image node - https://github.com/ianstormtaylor/slate/blob/main/packages/slate/src/transforms/text.ts#L436 
If there is some content in image caption pasting is working

When we pass selection into `insertNodes` this code will execute and it will try to insert new node into parent of nearest block (in fact into the Editor) - https://github.com/ianstormtaylor/slate/blob/main/packages/slate/src/transforms/node.ts#L196

I also updated slate version as there some changes in logic for inserting nodes/fragment. I think we need to keep slate version up to date.

Demo:


packages/slate-editor/src/extensions/paste-slate-content/lib/withSlatePasting.ts
```diff
@@ -38,3 +44,14 @@ function withoutSlateFragmentData(data: DataTransfer): DataTransfer {
     const dataMap = Object.fromEntries(types.map((type) => [type, data.getData(type)]));
     return createDataTransfer(dataMap);
 }
+
+function handlePastingIntoImageCaption(editor: Editor, fragment: SlateFragment) {
+    const [imageNode] = Array.from(Editor.nodes(editor, { match: isImageNode })).at(0) ?? [];
+
+    if (imageNode && EditorCommands.isNodeEmpty(editor, imageNode)) {
+        Transforms.insertNodes(editor, fragment, { at: editor.selection?.anchor.path });
```
aspirisen, line 52: Passing explicit path does the trick and this code is skipped https://github.com/ianstormtaylor/slate/blob/main/packages/slate/src/transforms/node.ts#L196 and therefore new node is inserted into image caption node

packages/slate-editor/src/modules/nodes-hierarchy/queries/isTextualNode.ts
```diff
@@ -0,0 +1,8 @@
+import { isHeadingNode, isParagraphNode } from '@prezly/slate-types';
+import type { Node } from 'slate';
+
+import { isInlineNode } from './isInlineNode';
+
+export function isTextualNode(node: Node) {
+    return isInlineNode(node) || isParagraphNode(node) || isHeadingNode(node);
```
e1himself, line 7: `isQuoteNode`?

e1himself, line 7: > `isInlineNode(node)`

Do we have to unwrap all inline nodes? What about placeholders? Right now it's possible to use them in the caption:

Screenshot from 2022-08-24 12-20-23

packages/slate-editor/src/extensions/paste-slate-content/lib/withSlatePasting.ts
e1himself: I dislike the fact that the generic code of `withSlatePasting` started to be dependent on our model -- specifically it's aware of our ImageNodes. That's a code smell to me.

Is there a way to avoid hard-coding it to use special logic just for images? Why images are special? Can we change it around by checking for `editor.isVoid` or `editor.isRichBlock`?

aspirisen: > Why images are special?
Because they can be technically empty as the children is empty, and at the same visually is not empty as there is an image. When you past smth into empty paragraphs it will be replaced with new content and it feels expected, but not for images.

I just noticed that we have the same situation for quotes, agree with general solution. Maybe we could add special callback into Extension and check it here?

e1himself: > Because they can be technically empty as the children is empty, and at the same visually is not empty as there is an image. When you past smth into empty paragraphs it will be replaced with new content and it feels expected, but not for images.

I believe that `editor.isRichBlock()` is a method to do exactly this check. Rich blocks, even having empty children, should not be considered empty.

packages/slate-editor/src/modules/nodes-hierarchy/queries/isTextualNode.ts
aspirisen: Well, it is an inline node, maybe there can be special cases that users will want to have placeholders. I am not sure that such cases will occur, but do not see a reason why should we forbid it.

e1himself: >  but do not see a reason why should we forbid it

But this change here will unwrap the pasted placeholders, as they are inline. Right?

aspirisen: unwrapping does not working for placeholder because it has not child elements

e1himself:
Reviewing now

packages/slate-editor/src/extensions/blockquote/BlockquoteExtension.tsx
```diff
@@ -20,6 +20,7 @@ export function BlockquoteExtension(): Extension {
                 BLOCKQUOTE: () => ({ type: QUOTE_NODE_TYPE }),
             }),
         },
+        isRichBlock: isQuoteNode,
```
e1himself, line 23: Hm... This is no good. We should not enable `isRichBlock` for blockquotes.

There's additional special formatting for rich blocks appearing in email campaigns. They are constrained to a 600px width column in the email rendering logic. Here's how a right-aligned quote looks now:

image

See #194 and MT-4987

e1himself: I guess we have no other choice, but to introduce a new check callback for this use-case. 

But instead of declaring this globally on extensions, we can pass it as a parameter to the `PasteSlateContentExtension` constructor:

```ts
// PasteSlateContentExtension.ts

/**
 * @param isPreservedBlock Defines which blocks should be preserved, if pasted empty content.
 */
export function PasteSlateContentExtension(
    isPreservedBlock: (editor: Editor, node: Node) => boolean = alwaysFalse,
): Extension {
    return {
        id: EXTENSION_ID,
        withOverrides: withSlatePasting(isPreservedBlock),
    };
}

```

```ts
// getEnabledExtensions.ts

yield PasteSlateContentExtension(
   (editor, node) => isQuoteNode(node) || editor.isRichBlock(node),
);
```

What would you say about this?

aspirisen: I thought that Extension params are more about configuration or providing connection to external api. Here i think that removing node when you paste content is a bug and i am not sure that it should be controlled outside

e1himself: > Here i think that removing node when you paste content is a bug

So do you suggest to make this behavior default for all types of nodes?

aspirisen: not to all, to those which have empty children (but can have them) and visually not empty. Maybe we should somehow define `isEmpty` behaviour? What do you think, if we have image but have no caption can we consider that this image node is not empty?

e1himself: Sceptical.

While providing this wider `isEmpty` check for images would make sense, it wouldn't make any sense for Quote nodes. As a quote without children is in fact empty. Treating it otherwise, would pose other problems, I am sure.

e1himself (approved):

PR #295, opened 2022-08-31
[DEV-8696] Feature - Editor Imperative API to insert nodes programmatically

e1himself:
- Provide an additional imperative API method: `insertNodes()`
- Tweak ContactNode card rendering: do not render description block if it's empty
- Drop `newsroomSettingsUrl` prop from all components
- Additionally pass `onClose` to `renderSearch`

### DIFF
--- a/packages/slate-editor/src/modules/editor/Editor.tsx
+++ b/packages/slate-editor/src/modules/editor/Editor.tsx
@@ -585,7 +585,6 @@ export const Editor = forwardRef<EditorRef, EditorProps>((props, forwardedRef) =
                         availableWidth={availableWidth}
                         containerRef={containerRef}
                         events={events}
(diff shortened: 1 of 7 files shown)


packages/slate-editor/src/modules/editor/Editor.tsx
```diff
@@ -582,7 +585,6 @@ export const Editor = forwardRef<EditorRef, EditorProps>((props, forwardedRef) =
                         availableWidth={availableWidth}
                         containerRef={containerRef}
                         events={events}
-                        newsroomSettingsUrl={withPressContacts.newsroomSettingsUrl}
```
e1himself, line 585: This prop wasn't really used here. It has been received from the external code and then passed as is to `renderSearch()`, which is external code too. The scheme can be simplified by letting the external code solve it without proxying the prop through the editor.

aspirisen (approved):

PR #328, opened 2022-10-03
[DEV-8897] Snippet insert animation

aspirisen:


packages/slate-editor/src/extensions/blink-nodes/components/BlinkNodes.tsx
```diff
@@ -0,0 +1,67 @@
...
+    top,
+    bottom,
+    containerRef,
+    clear,
+}: {
+    editor: Editor;
+    top: Node;
+    bottom: Node;
+    containerRef: RefObject<HTMLDivElement>;
+    clear: () => void;
```
e1himself, line 40: I suggest renaming this prop to have `on` prefix. To match the common callback prop convention. 

Also, `onComplete` seems to be more clear, IMO.

packages/slate-editor/src/extensions/blink-nodes/components/BlinkNodes.tsx
```diff
@@ -0,0 +1,67 @@
...
+        <>
+            {editor.nodesToBlink.map(([top, bottom], index) => (
+                <Blinker
+                    key={index}
+                    editor={editor}
+                    top={top}
+                    containerRef={containerRef}
+                    bottom={bottom}
+                    clear={() =>
+                        (editor.nodesToBlink = editor.nodesToBlink.filter((_, i) => i !== index))
```
e1himself, line 21: What if there's another blink pair added while this one is still animating?
Wouldn't it be better to check for object identity instead?

```ts
editor.nodesToBlink = editor.nodesToBlink.filter((pair) => pair[0] === top && pair[1] === bottom)
```
?

packages/slate-editor/src/extensions/snippet/lib/useFloatingSnippetInput.ts
```diff
@@ -42,8 +42,10 @@ export function useFloatingSnippetInput(editor: Editor): [State, Actions] {
 
         try {
             EditorCommands.insertNodes(editor, node.children);
+            editor.blink(node.children.at(0), node.children.at(-1));
```
e1himself, line 45: Have you considered using `Editor.rangeRef` here instead of node objects?

The downside of using nodes, is that if there's any change happens to them after insertion (including normalizations) -- the animation won't work.

packages/slate-editor/src/extensions/snippet/lib/useFloatingSnippetInput.ts
```diff
@@ -42,8 +42,10 @@ export function useFloatingSnippetInput(editor: Editor): [State, Actions] {
...
             savedSelection.restore(editor, { focus: true });
         } catch (error) {
+            console.error(error);
```
e1himself, line 48: Should it be there for the production code? (just checking if it's not something you've forgot to remove)

packages/slate-editor/src/extensions/blink-nodes/BlinkNodesExtension.tsx
```diff
@@ -0,0 +1,20 @@
+import type { Extension } from '@prezly/slate-commons';
+
+export function BlinkNodesExtension(): Extension {
```
e1himself, line 3: IMO "blink" is not precisely what it does. It's more like "highlight". 

Blinking is when it goes on and off, repeated. Not once.

e1himself (approved):
Well done man! :+1: 

I have a couple of minor non-blocking suggestions below. But otherwise, good job :+1:

packages/slate-editor/src/extensions/snippet/lib/useFloatingSnippetInput.ts
aspirisen: How rangeRef could help if the nodes are not inserted?

packages/slate-editor/src/extensions/blink-nodes/BlinkNodesExtension.tsx
aspirisen: what about "Flash" ?

PR #335, opened 2022-10-13
[DEV-9017] Feature - Enforce content structure based on the provided schema

kudlajz:
I've implemented an extension called `AllowedBlocks` (using suggested name, but maybe a better name could be more fitting based on the behaviour).

Please see [DEV-9008](https://linear.app/prezly/issue/DEV-9008/design-gallery-creationediting-aside) for explanation on why we need these changes.


packages/slate-editor/src/extensions/allowed-blocks/AllowedBlocksExtension.ts
```diff
@@ -0,0 +1,42 @@
...
+import { type Editor, type NodeEntry, Transforms } from 'slate';
+
+import type { AllowedBlocksExtensionConfiguration, Block } from './types';
+
+export const EXTENSION_ID = 'AllowedBlocksExtension';
+
+export function AllowedBlocksExtension({ blocks }: AllowedBlocksExtensionConfiguration): Extension {
+    return {
+        id: EXTENSION_ID,
+        normalizeNode(editor: Editor, [node, path]: NodeEntry) {
```
kudlajz, line 13: Here's an [example](https://github.com/prezly/prezly/pull/11617/files#diff-f9a6d5f0592725fbbd69b38ca49032ddafed7e0bf91f222668b4d6ec5897d169R42-R53) on how it's used.

Each entry in the `blocks` array corresponds to the top-level path (or row if you want).
For the current `path`, we check whether the current node is one of the types specified in `types` array and if it's not, it will be replaced by a block returned from `defaultBlockFn`.

packages/slate-editor/src/extensions/allowed-blocks/types.ts
```diff
@@ -0,0 +1,12 @@
+import type { ElementNode } from '@prezly/slate-types';
+
+type DefaultBlockFn = () => ElementNode;
+
+export interface Block {
+    types: string[];
+    defaultBlockFn: DefaultBlockFn;
```
kudlajz, line 7: I wasn't sure about the name. Maybe `getDefaultBlock` or something? 🤔

packages/slate-editor/src/extensions/allowed-blocks/AllowedBlocksExtension.ts
```diff
@@ -0,0 +1,42 @@
+import type { Extension } from '@prezly/slate-commons';
+import { isElementNode } from '@prezly/slate-types';
+import { get } from 'lodash-es';
+import { type Editor, type NodeEntry, Transforms } from 'slate';
+
+import type { AllowedBlocksExtensionConfiguration, Block } from './types';
+
+export const EXTENSION_ID = 'AllowedBlocksExtension';
+
+export function AllowedBlocksExtension({ blocks }: AllowedBlocksExtensionConfiguration): Extension {
```
kudlajz, line 10: Even though this extension's purpose right now is to normalize and enforce a certain content structure, I didn't want to call it a normalizing extension since the implemented behaviour is pretty "dumb" right now so anything more advanced will require thinking about this more.

packages/slate-editor/src/extensions/placeholders/components/PlaceholderElement.tsx
```diff
@@ -13,6 +13,7 @@ import { type Props as BaseProps, Placeholder } from './Placeholder';
 export type Props = RenderElementProps &
     Pick<BaseProps, 'icon' | 'title' | 'description' | 'format' | 'onClick' | 'onDrop'> & {
         element: PlaceholderNode;
+        removable: boolean;
```
kudlajz, line 16: I wasn't sure whether to make this optional and make it default to `true` since it's already being provided by the extension configuration.

e1himself (approved):

packages/slate-editor/src/extensions/allowed-blocks/AllowedBlocksExtension.ts
kudlajz: Hmm, this looks really nice!

PR #336, opened 2022-10-14
Fix - Improve behaviour of image/gallery blocks

kudlajz:
When editing an image or a gallery, the original element is removed first, replaced by an empty paragraph, then replaced by the loader element and then replaced by the actual element.

This makes it fairly difficult to use the new `AllowedBlocks` extension because when you edit a gallery/image, for a brief moment the original element will be replaced by a paragraph. If you don't allow paragraphs, it will replace it with a custom block, which is not the desired outcome.

With these changes, when editing, the element is not removed and there's no paragraph used as an intermediary medium. Instead the element is replaced by a loader which is then replaced by the actual updated element.

There's a new `mode` option that affects whether to use replace or simply insert nodes as in some cases (e.g. `withFilePasting`) we don't want to actually replace the nodes but insert new ones.

I have tested the new behaviour and everything seems to work the same as it currently does in production.

Thanks to these changes I was also able to fix two bugs:
1. when you edited a gallery, it would place an empty paragraph above the gallery. That doesn't seem useful.
2. when you edited a gallery, it would focus the empty paragraph below it instead of keeping the focus on the gallery (that's how it works for images)

We still create a new paragraph below the element when creating it.

### DIFF
--- a/packages/slate-editor/src/modules/editor/lib/insertUploadingFile.ts
+++ b/packages/slate-editor/src/modules/editor/lib/insertUploadingFile.ts
@@ -1,6 +1,7 @@
 import type { ProgressPromise } from '@prezly/progress-promise';
 import { EditorCommands } from '@prezly/slate-commons';
-import type { Editor, Element } from 'slate';
+import type { Editor, Element} from 'slate';
+import { Transforms } from 'slate';
 
 import type { LoaderContentType } from '#extensions/loader';
 import {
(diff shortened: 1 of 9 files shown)


packages/slate-editor/src/modules/editor/lib/insertUploadingFile.ts
```diff
@@ -29,17 +30,25 @@ export async function insertUploadingFile<T>(
...
         message: loaderMessage,
     });
-    EditorCommands.insertNodes(editor, [loader], { ensureEmptyParagraphAfter });
+
+    if (mode === 'insert') {
+        EditorCommands.insertNodes(editor, [loader], { ensureEmptyParagraphAfter: true });
+    }
+
+    if (mode === 'replace') {
+        Transforms.setNodes(editor, loader);
```
aspirisen, line 49: jfyi, as i remember setNodes will not set children, do not know if it affects something, but think it is worth to mention here

PR #337, opened 2022-10-17
Feature - Allow providing popper options to control how the element menu behaves

kudlajz:
When implementing the [media gallery aside](https://linear.app/prezly/issue/DEV-9008/design-gallery-creationediting-aside), we needed to display the menu on the left side as there's more space than on the right.

However the menu was not displaying nicely as the aside's content is a scrollable container, therefore I've tweaked the Popper a little bit so now it follows the user when scrolling as long as the element is still in view.

You can see it in action here (video uploaded to Slack):
- [media gallery aside](https://prezly.slack.com/files/U7HKXUECF/F046UQ9NYQ2/gallery_scrolling.mp4)
- [story editor](https://prezly.slack.com/files/U7HKXUECF/F046E82KPDM/story_scrolling.mp4)

I recommend hiding whitespaces when reviewing.


packages/slate-editor/src/components/EditorBlock/Menu.tsx
```diff
@@ -6,78 +7,93 @@ import { Popper } from 'react-popper';
...
     children: ReactNode;
     className?: string;
     onClick?: (event: MouseEvent) => void;
+    popperOptions: PopperOptionsContextType;
     reference: HTMLElement;
 }
 
-const VIRTUAL_REFERENCE_SIZE = 42;
+const TETHER_OFFSET_AMOUNT = 4;
+const TETHER_OFFSET_OUTLINE_SIZE = 6;
```
kudlajz, line 23: I've made the menu align to the outline, currently it was a few pixels off (somewhere between the element and its outline).

I quite like it but wasn't sure if we want it or if it should be aligned to the element rather than the outline, so let me know!

kudlajz, line 23: I've made the menu align to the outline, currently it was a few pixels off (somewhere between the element and its outline).

I quite like it but wasn't sure if we want it or if it should be aligned to the element rather than the outline, so let me know!

packages/slate-editor/src/modules/popper-options-context/types.ts
```diff
@@ -0,0 +1,8 @@
+import type { PreventOverflowModifier } from '@popperjs/core/lib/modifiers/preventOverflow';
+
+export interface Context {
+    modifiers?: {
+        preventOverflow?: PreventOverflowModifier['options'];
+    };
+    placement?: 'left-start' | 'right-start';
```
aspirisen, line 7: Is it possible to make it automatic, if there is not place on the right move it to the left, probably there should be such props for popper

packages/slate-editor/src/modules/editor/Editor.tsx
```diff
@@ -529,216 +531,218 @@ export const Editor = forwardRef<EditorRef, EditorProps>((props, forwardedRef) =
...
-                     * children are inline or text nodes they will be removed. This ensures that there
-                     * are always block nodes in the editor so that behaviors like "splitting a block
-                     * in two" work as expected.
-                     */
-                    onChange(newValue as Element[]);
-                    variables.onChange(editor);
-                    userMentions.onChange(editor);
-                }}
-                value={initialValue}
+        <PopperOptionsContext.Provider value={popperMenuOptions}>
```
kudlajz, line 534: I couldn't think of a better way to achieve this without prop drilling.

packages/slate-editor/src/modules/popper-options-context/types.ts
kudlajz: Yes, it is possible but it doesn't work well in the story editor because when the toolbox goes to the left, it's covered by the sidebar due to z-index issues there.

It would require more work to make this work nicely so I've opted for this simpler approach, but it's definitely doable.

aspirisen: I have some suggestion to make the menu adaptive. Feel free to apply these changes in some way or just cancel that PR https://github.com/prezly/slate/pull/338

kudlajz:
@aspirisen Can you please approve if it's okay? I would like to get this released 🙂

aspirisen (approved):

PR #338, opened 2022-10-17
Suggestion - Auto menu placement

aspirisen:
Suggestion to make menu adaptive to the screen size

### DIFF
--- a/packages/slate-editor/src/components/EditorBlock/Menu.module.scss
+++ b/packages/slate-editor/src/components/EditorBlock/Menu.module.scss
@@ -1,7 +1,10 @@
+@import "styles/variables";
+
 .menu {
     width: 280px;
     position: absolute;
     backdrop-filter: blur(6px);
+    z-index: $toolbar-z-index;
 }
 
 .arrow {
--- a/packages/slate-editor/src/components/EditorBlock/Menu.tsx
+++ b/packages/slate-editor/src/components/EditorBlock/Menu.tsx
@@ -128,7 +128,10 @@ export class Menu extends Component<Props> {
                             contentEditable={false}
                             className={classNames(className, styles.menu)}
                             ref={ref}
-                            style={style}
+                            style={{
+                                ...style,
+                                zIndex: popperOptions.zIndex,
+                            }}
                             onClick={onClick}
                         >
                             <div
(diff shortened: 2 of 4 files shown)


packages/slate-editor/src/components/EditorBlock/Menu.tsx
```diff
@@ -108,37 +112,39 @@ export class Menu extends Component<Props> {
...
+                                    [styles.top]: placement.indexOf('top') >= 0,
+                                    [styles.bottom]: placement.indexOf('bottom') >= 0,
+                                    [styles.left]: placement.indexOf('left') >= 0,
+                                    [styles.right]: placement.indexOf('right') >= 0,
+                                })}
+                                {...arrowProps}
+                            />
+                            {children}
+                        </Toolbox.Panel>,
+                        document.getElementById('full-page-react-application') as HTMLElement,
```
aspirisen, line 145: There should be some container from props

kudlajz: I think we could add the prop into the context I've introduced.

packages/slate-editor/src/components/EditorBlock/Menu.module.scss
```diff
@@ -1,7 +1,10 @@
+@import "styles/variables";
+
 .menu {
     width: 280px;
     position: absolute;
     backdrop-filter: blur(6px);
+    z-index: $toolbar-z-index;
```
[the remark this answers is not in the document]
aspirisen: yeah, it looks like the best place, or just explicit prop on the editor to control editor styles (maybe we will have some cases where we need to control styles as well). 
It is up to you, i have no strong opinion on how to provide this z-index

packages/slate-editor/src/components/EditorBlock/Menu.tsx
kudlajz: I was able to make it work just by providing `document.body` here. Maybe we can use that for now and add the prop in the future when/if needed?

kudlajz (approved):
I've used `document.body` for the portal and added support for settings the z-index via the popper options.

packages/slate-editor/src/components/EditorBlock/Menu.tsx
aspirisen: Yeah, let's leave for now. The one problem could be if user's browser extensions will mutate body's elements, but it is an edge case for now

PR #339, opened 2022-10-18
[QRF-349] Remove unused embed props

aspirisen:
Looks like video prop is not used anymore
https://linear.app/prezly/issue/QRF-349/story-editor-embed-option-not-working


packages/slate-editor/src/modules/editor/types.ts
```diff
@@ -99,7 +99,6 @@ export interface EditorProps {
                   embed?: boolean;
                   link?: boolean;
```
e1himself, line 100: AFAIK, `link` shouldn't be used too. Please check.

e1himself (approved):

packages/slate-editor/src/modules/editor/types.ts
aspirisen: Looks like all of these props are unused

PR #340, opened 2022-10-18
Feature - Add blur method to the Editor component

kudlajz:
We want to be able to unfocus (blur) the editor in some cases.

### DIFF
--- a/packages/slate-commons/src/commands/blur.ts
+++ b/packages/slate-commons/src/commands/blur.ts
@@ -1,0 +1,6 @@
+import type { Editor } from 'slate';
+import { ReactEditor } from 'slate-react';
+
+export function blur(editor: Editor & ReactEditor): void {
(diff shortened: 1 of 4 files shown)


packages/slate-commons/src/commands/blur.ts
```diff
@@ -0,0 +1,6 @@
+import type { Editor } from 'slate';
+import { ReactEditor } from 'slate-react';
+
+export function blur(editor: Editor & ReactEditor): void {
+    ReactEditor.deselect(editor);
```
kudlajz, line 5: For some reason `ReactEditor.blur` doesn't do what I'd expect but `ReactEditor.deselect` does.

I wasn't sure about the function name but I think the `blur` is still correct as it does what is expected.

aspirisen (approved):

PR #343, opened 2022-10-24
[QRF-13] Fix image focusing while typing in the captions

aspirisen:
https://linear.app/prezly/issue/QRF-13/image-caption-editing-is-hard-when-image-height-is-bigger-than-the

### DIFF
--- a/packages/slate-editor/src/extensions/image/components/ImageElement.tsx
+++ b/packages/slate-editor/src/extensions/image/components/ImageElement.tsx
@@ -100,9 +100,12 @@ export function ImageElement({
         [withSizeOptions, element.file],
     );
 
+    const { ref, ...slateAttributes } = attributes;
+
     return (
         <ResizableEditorBlock
-            {...attributes}
+            {...slateAttributes}
+            ref={isSupportingCaptions ? undefined : ref}
             align={align}
             element={element}
             layout={layout}
@@ -112,6 +115,7 @@ export function ImageElement({
             renderBelowFrame={
                 isSupportingCaptions && (
                     <div
+                        ref={ref}
                         className={classNames(styles.Caption, {
                             [styles.empty]: isCaptionEmpty,
                             [styles.withPlaceholder]: isCaptionPlaceholderVisible,
--- a/packages/slate-editor/src/extensions/image/components/ImageElement.tsx
+++ b/packages/slate-editor/src/extensions/image/components/ImageElement.tsx
@@ -100,12 +100,9 @@ export function ImageElement({
         [withSizeOptions, element.file],
     );
 
-    const { ref, ...slateAttributes } = attributes;
-
     return (
         <ResizableEditorBlock
-            {...slateAttributes}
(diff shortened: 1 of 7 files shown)


packages/slate-editor/src/extensions/image/components/ImageElement.tsx
```diff
@@ -100,9 +100,12 @@ export function ImageElement({
         [withSizeOptions, element.file],
     );
 
+    const { ref, ...slateAttributes } = attributes;
+
     return (
         <ResizableEditorBlock
-            {...attributes}
+            {...slateAttributes}
+            ref={isSupportingCaptions ? undefined : ref}
```
e1himself, line 108: Nope. The given `ref` is required to be set on the top-level DOM node for the element. Otherwise a lot of internal logic inside Slate just breaks.

image

e1himself (changes requested):

packages/slate-editor/src/extensions/image/components/ImageElement.tsx
aspirisen: Do you have some other ideas how that can be fixed?

e1himself: I'd try to look into two things that might be causing this, or might provide a way to fix this:

- Our homebrew `useCursorInView` hook
- Slate native `scrollSelectionIntoView` property

aspirisen: Looks like `useCursorInView `. Thanks for pointing the direction ❤️

packages/slate-editor/src/modules/editor/lib/useCursorInView.ts
```diff
@@ -29,7 +30,7 @@ function ensureCursorInView(editor: Editor, parameters: Parameters): void {
     }
     const [currentNode] = EditorCommands.getCurrentNodeEntry(editor) || [];
 
-    if (Range.isExpanded(editor.selection) && !isImageNode(currentNode)) {
+    if (Range.isExpanded(editor.selection) || isImageNode(currentNode)) {
```
aspirisen, line 33: This is opposite to these changes https://github.com/prezly/slate/pull/149/files#diff-3189cee2b014dfe319eea38ebdec2dd4e9d0447af3542754a6f35aeb7cdef1aeR28 what is weird.

@OskarGasiorowski could you explain what behavior you tried to fix in that PR for images?

e1himself (approved):

packages/slate-editor/src/modules/editor/lib/useCursorInView.ts
```diff
@@ -18,6 +18,7 @@ export function useCursorInView(editor: Editor, parameters: false | Parameters =
         try {
             ensureCursorInView(editor, parameters);
         } catch (error) {
+            console.error(error);
```
e1himself, line 21: Is this necessary?

PR #345, opened 2022-10-25
[QRF-151] Allow block auto-format only in paragraphs

aspirisen:
https://linear.app/prezly/issue/QRF-151/editor-fails-using-in-unordered-lists

### DIFF
--- a/packages/slate-editor/src/extensions/autoformat/withAutoformat.ts
+++ b/packages/slate-editor/src/extensions/autoformat/withAutoformat.ts
@@ -1,4 +1,5 @@
 import { EditorCommands } from '@prezly/slate-commons';
+import { isParagraphNode } from '@prezly/slate-types';
 import type { Editor } from 'slate';
 
 import { autoformatBlock, autoformatMark, autoformatText } from './transforms';
@@ -28,6 +29,12 @@ export function withAutoformat<T extends Editor>(editor: T, rules: AutoformatRul
             if (query && !query(editor, { ...rule, text })) continue;
 
             if (editor.selection) {
+                const [currentNode] = EditorCommands.getCurrentNodeEntry(editor) ?? [];
+
+                if (mode === 'block' && !isParagraphNode(currentNode)) {
+                    return;
+                }
+
                 const formatter = autoformatters[mode];
 
                 const formatResult = formatter?.(editor, {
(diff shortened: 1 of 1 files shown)


packages/slate-editor/src/extensions/autoformat/withAutoformat.ts
```diff
@@ -28,6 +29,12 @@ export function withAutoformat<T extends Editor>(editor: T, rules: AutoformatRul
             if (query && !query(editor, { ...rule, text })) continue;
 
             if (editor.selection) {
+                const [currentNode] = EditorCommands.getCurrentNodeEntry(editor) ?? [];
+
+                if (mode === 'block' && !isParagraphNode(currentNode)) {
+                    return;
+                }
```
e1himself, line 36: How about using bold and italic formatting shortcuts in Quotes, Lists, or Headings?

As I've said before, I don't believe this is a good fix.

e1himself (changes requested):
As I've already mentioned before, I don't like this fix. I'd expect the formatting shortcuts work in headings too.
But alright, let's see if people complain.

packages/slate-editor/src/extensions/autoformat/withAutoformat.ts
e1himself: Alright. After reading the code more and testing more, I've realized that the intention was to keep using mark and text auto-formatters without limits. But there's a bug. It should be `continue` instead of `return` here.

packages/slate-editor/src/extensions/autoformat/withAutoformat.ts
```diff
@@ -28,6 +29,12 @@ export function withAutoformat<T extends Editor>(editor: T, rules: AutoformatRul
             if (query && !query(editor, { ...rule, text })) continue;
 
             if (editor.selection) {
+                const [currentNode] = EditorCommands.getCurrentNodeEntry(editor) ?? [];
+
+                if (mode === 'block' && !isParagraphNode(currentNode)) {
+                    return;
```
e1himself, line 35:
```suggestion
                    continue;
```

e1himself (changes requested):

e1himself (approved):

e1himself (approved):
I still don't like that the formatting shortcuts will not work in headings. But alright. Let's see if someone else complains.

PR #346, opened 2022-10-25
[QRF-300] Fix implicit alignment prop bypassing when image is removed

aspirisen:
https://linear.app/prezly/issue/QRF-300/weird-arrow-keys-and-image-behaviour-in-the-editor

### DIFF
--- a/packages/slate-editor/src/extensions/image/ImageExtension.tsx
+++ b/packages/slate-editor/src/extensions/image/ImageExtension.tsx
@@ -1,12 +1,13 @@
 import type { Extension } from '@prezly/slate-commons';
 import { createDeserializeElement, EditorCommands, withoutNodes } from '@prezly/slate-commons';
 import type { ImageNode, ParagraphNode } from '@prezly/slate-types';
+import { Alignment } from '@prezly/slate-types';
 import { IMAGE_NODE_TYPE, isImageNode } from '@prezly/slate-types';
 import { isHotkey } from 'is-hotkey';
 import { noop } from 'lodash-es';
 import type { KeyboardEvent } from 'react';
 import React from 'react';
-import type { Editor } from 'slate';
+import type { Editor, NodeEntry, Node } from 'slate';
 import { Path, Transforms } from 'slate';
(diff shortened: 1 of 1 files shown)


packages/slate-editor/src/extensions/image/ImageExtension.tsx
```diff
@@ -159,3 +154,14 @@ export const ImageExtension = ({
     serialize: (nodes) => withoutNodes(nodes, isImageCandidateElement),
     withOverrides: withImages,
 });
+
+function replaceImageWithParagraph(editor: Editor, nodeEntry: NodeEntry<Node>) {
+    Transforms.setNodes<ImageNode | ParagraphNode>(
+        editor,
+        createParagraph({ align: Alignment.LEFT }),
```
e1himself, line 161: I think it'd be better to remove the `align` prop, instead of forcing it to `left`.

The reason is that, if omitted, paragraphs will inherit editor global `align` prop alignment (which may be different from `left`; i.e. RTL locales have it set to `right`).

So:

```tsx
Transforms.setNodes<ImageNode | ParagraphNode>(editor, createParagraph(), {
    at: nodeEntry[1],
    match: isImageNode,
});
Transforms.unsetNodes<ParagraphNode>(editor, 'align', {
    at: nodeEntry[1], 
    match: isParagraphNode,
});
```

packages/slate-commons/src/commands/replaceNode.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+interface Props<T extends Node> {
+    entry: NodeEntry<T>;
+    match: NodeMatch<T>;
+}
+
+export function replaceNode<Original extends Node, New extends Node>(
+    editor: Editor,
+    props: Props<Original>,
+    createNode: () => New,
```
e1himself, line 12: I guess it would make more sense to pass the actual new node here (not a factory).

```ts
export function replaceNode<Original extends Node, New extends Node>(
    editor: Editor,
    props: Props<Original>,
    node: Omit<New, 'children'>,
)

e1himself (approved):

PR #347, opened 2022-10-26
[QRF-389] Allow auto-placement for toolbox

aspirisen:
https://linear.app/prezly/issue/QRF-389/make-editor-toolbox-to-have-automatic-placement


packages/slate-editor/src/modules/popper-options-context/types.ts
```diff
@@ -1,9 +1,11 @@
 import type { PreventOverflowModifier } from '@popperjs/core/lib/modifiers/preventOverflow';
+import type { RefObject } from 'react';
 
 export interface Context {
     modifiers?: {
         preventOverflow?: PreventOverflowModifier['options'];
     };
     placement?: 'left-start' | 'right-start';
+    autoPlacementPortal?: RefObject<HTMLElement | undefined>;
```
kudlajz, line 9: Would it be better to split this into two props?

I could imagine someone wanting just to use the portal and someone using both the auto placement and portal.

Maybe you could add `auto` as a possible option to `placement`? Then, if `placement === auto`, you can use `allowedAutoPlacements` property to specify `right-start` and `left-start`.

kudlajz: @aspirisen I see you've asked for my review again but I've left a comment here yesterday. Can you check it out?

aspirisen: Sorry, forgot to push :)

kudlajz (approved):
Looks good 👍 I just hope it doesn't break the scrolling behaviour 🙂

e1himself:
This is minor, but I've noticed that there's a regression in arrow positioning.

This is what I have in `main` locally:

image

This is [the spec](https://www.figma.com/file/KD30VWNjFh6mVWkg6Taqhe/Editor%3A-Story%2C-Campaign-%26-Pitch?node-id=2466%3A163475):

image

aspirisen:
@e1himself this is on production, it looks like this regression is out of scope of this pr?

e1himself:
@aspirisen true. Maybe @kudlajz has introduced it with #337?

kudlajz:
Yes, you're right. It was introduced there. I will check with UX people whether the arrow should still be fixed at the top considering the menu now moves when scrolling 👍

e1himself:
>  considering the menu now moves when scrolling

Just a note, that the original specs for the menu behavior already required the menu to move up/down when the page is being scrolled.

image

PR #342, opened 2022-10-21
[QRF-288] Rework resize handles

aspirisen:
https://linear.app/prezly/issue/QRF-288/image-settings-menu-covering-resize-button

PR #351, opened 2022-11-01
[QRF-288] Fix - Non-movable image resize handles

e1himself:
Followup on https://github.com/prezly/slate/pull/342

Hide non-movable image resize handles for left- or right-aligned images. As using them to resize  feelss really weird, as they do not move.


aspirisen:
Hm, for me this looks weird, this handlers need to resize image, it doesn't depend on the image position

aspirisen (approved):

PR #352, opened 2022-11-01
[DEV-6123] Allow optional Shift when detecting trigger character in FloatingAddMenu

kudlajz:
Some non-English keyboard layouts have to use Shift with a different key to produce the `/` trigger character. The code checking this event was too strict and did not allow the Shift key.

### DIFF
--- a/packages/slate-editor/src/extensions/floating-add-menu/FloatingAddMenuExtension.tsx
+++ b/packages/slate-editor/src/extensions/floating-add-menu/FloatingAddMenuExtension.tsx
@@ -5,8 +5,8 @@ import type { KeyboardEvent } from 'react';
 import { isMenuHotkey, MENU_TRIGGER_CHARACTERS, shouldShowMenuButton } from './lib';
 
 function isTriggerInput(event: KeyboardEvent) {
-    return MENU_TRIGGER_CHARACTERS.some((tiggerKey) =>
-        isHotkey(tiggerKey, { byKey: true }, event.nativeEvent),
+    return MENU_TRIGGER_CHARACTERS.some((triggerKey) =>
+        isHotkey(`shift?+${triggerKey}`, { byKey: true })(event),
     );
 }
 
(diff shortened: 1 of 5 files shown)


packages/slate-editor/src/extensions/floating-add-menu/FloatingAddMenuExtension.tsx
```diff
@@ -1,12 +1,11 @@
...
+import { isHotkey, type KeyboardEventLike } from 'is-hotkey';
 
 import { isMenuHotkey, MENU_TRIGGER_CHARACTERS, shouldShowMenuButton } from './lib';
 
-function isTriggerInput(event: KeyboardEvent) {
-    return MENU_TRIGGER_CHARACTERS.some((tiggerKey) =>
-        isHotkey(tiggerKey, { byKey: true }, event.nativeEvent),
+function isTriggerInput(event: KeyboardEventLike) {
+    return MENU_TRIGGER_CHARACTERS.some((triggerKey) =>
+        isHotkey(`shift?+${triggerKey}`, { byKey: true })(event),
```
aspirisen, line 8: In that case probably better to save `isHotkey(`shift?+${triggerKey}`, { byKey: true })` in order not to create the function every time

OskarGasiorowski (approved):

packages/slate-editor/src/extensions/floating-add-menu/FloatingAddMenuExtension.tsx
kudlajz: I have reworked the hotkey detection because it was using two identical characters to make the check. I've noticed that @OskarGasiorowski fixed it in https://github.com/prezly/slate/pull/146 but he also enabled the `byKey` option which actually fixes the issue. Adding the "second" trigger character was completely useless since it was the same character anyway 😄 

From the package description itself
> By default the hotkey string is checked using event.which. But you can also pass in byKey: true to compare using the [KeyboardEvent.key](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key) API, which stays the same regardless of keyboard layout.

So, when using `byKey` it doesn't matter if you press the regular slash on US keyboard, shift+something on non-US keyboard or numpad divide. `event.key` will always be `"/"`.

I've also defined the functions inside the module directly so it won't be recreated on every call.

See https://github.com/prezly/slate/pull/352/commits/25e89ae5f6ee7da8f3754aa6b93d629778ff1472.

cc @e1himself @aspirisen

PR #353, opened 2022-11-07
[QRF-163] Handle pasting content with "new line" breaks

aspirisen:
https://linear.app/prezly/issue/QRF-163/pitch-editor-inserts-spacing-when-pasting-content

### DIFF
--- a/packages/slate-editor/src/extensions/paragraphs/ParagraphsExtension.tsx
+++ b/packages/slate-editor/src/extensions/paragraphs/ParagraphsExtension.tsx
@@ -8,6 +8,7 @@ import { composeElementDeserializer } from '#modules/html-deserialization';
 
 import { ParagraphElement } from './components';
 import {
+    createParagraph,
     normalizeRedundantParagraphAttributes,
     normalizeUnknownElement,
     parseSerializedElement,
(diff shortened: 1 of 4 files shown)


aspirisen:
@e1himself could you take a look

e1himself (commented):
I've tried to reproduce the scenario described in #QRF-163, and it's still happening.

Also, it looks like it became even worse. Please see the attached video.

[Screencast from 11-15-2022 12:34:32 PM.webm](https://user-images.githubusercontent.com/370680/201898146-c6293d9a-ca62-41a8-b72e-fd1be6d337f9.webm)

aspirisen:
@e1himself looks like it is Firefox specific behavior (in case if you copy from ff). Chrome and FF provide different markup when you copy content. For example if you copy from chrome and paste in FF the result is good. The example above is not caused by this changes as the same we have on prod

e1himself:
@aspirisen I don't understand. Do you suggest to fix the exact problem described in QRF-163 for Chrome only and ignore Firefox?

e1himself:
I still have the problem in Chrome (but less pronounced, as in Firefox): 

[Screencast from 11-21-2022 02:53:08 PM.webm](https://user-images.githubusercontent.com/370680/203060284-34acdabe-4be3-4604-bc49-c1085d3a6e06.webm)

PR #355, opened 2022-11-18
[QRF-119] Forbid '/' in file names

aspirisen:
https://linear.app/prezly/issue/QRF-119/unable-to-load-attached-pdf-in-campaign-email

### DIFF
--- a/packages/slate-editor/src/modules/editor/lib/file-attachment/handleEditAttachment.ts
+++ b/packages/slate-editor/src/modules/editor/lib/file-attachment/handleEditAttachment.ts
@@ -16,7 +16,7 @@ export async function handleEditAttachment(editor: Editor, element: Partial<Atta
     }
 
     const newElement = mapFilename({ ...currentFileAttachment, ...element }, (filename) =>
-        filename.replaceAll('/', ':'),
+        filename.replaceAll('/', '_'),
     );
 
     Transforms.setNodes<AttachmentNode>(editor, newElement, {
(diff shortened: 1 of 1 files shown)


kudlajz (approved):

packages/slate-editor/src/modules/editor/lib/file-attachment/handleEditAttachment.ts
```diff
@@ -15,16 +15,18 @@ export async function handleEditAttachment(editor: Editor, element: Partial<Atta
         return;
     }
 
-    Transforms.setNodes<AttachmentNode>(editor, element, {
+    const newElement = { ...currentFileAttachment, ...element };
+
+    newElement.file.filename = newElement.file.filename.replaceAll('/', ':');
```
e1himself, line 20: 1. Should we maybe use `_` instead of `:` ? It's a safer choice for a filename IMO.
2. I have a bad feeling about mutating `file` property object. This can lead to something we do not expect. 
  Should we maybe use a safer update strategy? 


```tsx
export async function handleEditAttachment(editor: Editor, update: Partial<AttachmentNode>) {
   // ...
   const element = mapFilename(
      { ...currentFileAttachment, ...update },
      (filename) => filename.replaceAll('/', '_'),
   );
   // ...
}

function mapFilename(element: AttachmentNode, map: (filename: string) => string): AttachmentNode {
   return {
      ...element,
     file: {
        ...element.file,
        filename: map(element.file.filename),
     }
   };
}
```

PR #356, opened 2022-11-21
[QRF-45] Fix horizontal story bookmark provider width

aspirisen:
https://linear.app/prezly/issue/QRF-45/story-cards-cutting-off-title
Before:
image

After:
image


kudlajz (approved):
Would have been nice to have some screenshots (before and after). Also the PR title makes no sense 🙂

PR #361, opened 2022-12-13
[QRF-439] Add option to lose focus when clicking outside of the editor

oleg-semyonov:
A side-effect of this change was that closing the menu via the X button also loses focus to the block.

In [QRF-441](https://linear.app/prezly/issue/QRF-441/media-galleries-edit-aside-component-focus-behaviour-is-wrong) we agreed that X button can stay with the current behavior.

Here's what it looks like in this PR:

https://user-images.githubusercontent.com/1641218/207571909-f846f8ad-a597-4e7c-bcbb-f6fa2c6a20c8.mov

### DIFF
--- a/packages/slate-editor/src/modules/editor/Editor.tsx
+++ b/packages/slate-editor/src/modules/editor/Editor.tsx
@@ -85,7 +85,7 @@ export const Editor = forwardRef<EditorRef, EditorProps>((props, forwardedRef) =
         decorate,
         id,
         initialValue,
-        loseFocusOnOutsideClick = true,
+        loseFocusOnOutsideClick = false,
         onIsOperationPendingChange,
         onKeyDown = noop,
         placeholder,
--- a/packages/slate-editor/src/modules/editor/Editor.tsx
+++ b/packages/slate-editor/src/modules/editor/Editor.tsx
@@ -85,7 +85,7 @@ export const Editor = forwardRef<EditorRef, EditorProps>((props, forwardedRef) =
         decorate,
         id,
         initialValue,
-        loseFocusOnOutsideClick = false,
+        blurOnOutsideClick = false,
         onIsOperationPendingChange,
         onKeyDown = noop,
         placeholder,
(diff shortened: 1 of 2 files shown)


packages/slate-editor/src/modules/editor/Editor.tsx
```diff
@@ -191,6 +192,35 @@ export const Editor = forwardRef<EditorRef, EditorProps>((props, forwardedRef) =
...
+
+            const isWithinMenuPortal = isTargetWithin(popperMenuOptions.portalNode?.current);
+            const isWithinEditor = isTargetWithin(containerRef.current);
+
+            if (!isWithinEditor && !isWithinMenuPortal) {
+                EditorCommands.blur(editor);
+            }
+        }
+
+        if (loseFocusOnOutsideClick) {
```
kudlajz, line 215: Hmm, if only there was a word for the opposite of focus 😋 How about.. `blurOnOutsideClick`? 😄

kudlajz (approved):
Looks good although I'm still not sure on bluring the editor when you close the toolbox. Probably not a big deal for mouse users but others who rely on focus could have troubles here I think.

packages/slate-editor/src/modules/editor/Editor.tsx
oleg-semyonov: Right! That's much better, thanks Lukas! ❤️

oleg-semyonov:
> Looks good although I'm still not sure on blurring the editor when you close the toolbox

When closing through keyboard, focus should stay, as this new prop only listens for `click` events

aspirisen (approved):
we have react-use library and looks like there is a hook for that https://github.com/streamich/react-use/blob/master/docs/useClickAway.md
maybe it is worth to try if it would fit here

oleg-semyonov:
> we have react-use library and looks like there is a hook for that https://github.com/streamich/react-use/blob/master/docs/useClickAway.md
maybe it is worth to try if it would fit here

Not sure if it would fit this use case, since I need to check for two components simultaneously (otherwise the editor will lose focus when clicking on the menu itself)

PR #364, opened 2022-12-19
Ignore `blurOnOutsideClick` events on textbox elements

oleg-semyonov:
This is a follow-up for #361. It turned out that while clicking anywhere outside of the story composer worked perfectly, clicking on any textbox element (e.g. Title and Subtitle textareas) would result in those elements losing focus as well. With this workaround, and additional logic on the app side, everything now works as intended. I've added extra info in the typedoc to warn about the new behavior.

### DIFF
--- a/packages/slate-editor/src/modules/editor/Editor.tsx
+++ b/packages/slate-editor/src/modules/editor/Editor.tsx
@@ -206,8 +206,11 @@ export const Editor = forwardRef<EditorRef, EditorProps>((props, forwardedRef) =
 
             const isWithinMenuPortal = isTargetWithin(popperMenuOptions.portalNode?.current);
             const isWithinEditor = isTargetWithin(containerRef.current);
+            const isTextboxElement =
+                clickTarget.tagName.toLowerCase() === 'textarea' ||
+                clickTarget.tagName.toLowerCase() === 'input';
 
-            if (!isWithinEditor && !isWithinMenuPortal) {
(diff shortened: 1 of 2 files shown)


packages/slate-editor/src/modules/editor/Editor.tsx
```diff
@@ -206,8 +206,11 @@ export const Editor = forwardRef<EditorRef, EditorProps>((props, forwardedRef) =
 
             const isWithinMenuPortal = isTargetWithin(popperMenuOptions.portalNode?.current);
             const isWithinEditor = isTargetWithin(containerRef.current);
+            const isTextboxElement =
+                clickTarget.tagName.toLowerCase() === 'textarea' ||
+                clickTarget.tagName.toLowerCase() === 'input';
 
-            if (!isWithinEditor && !isWithinMenuPortal) {
+            if (!isWithinEditor && !isWithinMenuPortal && !isTextboxElement) {
```
aspirisen, line 213: Is it possible to attach event not to `body` but `containerRef` ?

aspirisen (approved):

packages/slate-editor/src/modules/editor/Editor.tsx
oleg-semyonov: Exactly :)

e1himself (approved):
Doesn't look super reliable, but alright :man_shrugging:

PR #368, opened 2023-01-17
[DEV-9820] Change - ContactNode interface to allow inline contacts

e1himself:
- `ContactNode` interface has been changed to allow inline contacts
  - `id` and `uuid` props dropped from the `PressContact` interface
  - `PressContact` interface renamed to `ContactInfo`
  - Added `reference` property to the `ContactNode` interface


packages/slate-types/src/nodes/ContactNode.ts
```diff
@@ -7,8 +7,9 @@ export const CONTACT_NODE_TYPE = 'contact';
 
 export interface ContactNode extends ElementNode {
     type: typeof CONTACT_NODE_TYPE;
-    contact: PressContact;
     uuid: string;
+    reference: ContactNode['uuid'] | null;
```
kudlajz, line 11: Isn't it better to do `string | null` instead of the type referencing itself? 🤔 `uuid` and `reference` are supposed to be independent of each other anyway, right?

kudlajz (approved):

packages/slate-types/src/nodes/ContactNode.ts
e1himself: Oh, good catch! 

So it actually had to be `reference: NewsroomContact['uuid']`. 

Fixed with be2421afdcc387436960b391c9b901bd5f3441f3

PR #372, opened 2023-01-18
[DEV-9750] Inline Contacts extension

kudlajz:


packages/slate-editor/src/extensions/inline-contacts/InlineContactsExtension.tsx
```diff
@@ -0,0 +1,48 @@
...
+import { CONTACT_NODE_TYPE, isContactNode } from '@prezly/slate-types';
+import React from 'react';
+
+import { composeElementDeserializer } from '#modules/html-deserialization';
+
+import {
+    normalizeContactInfoAttributes,
+    normalizeContactNodeAttributes,
+    parseSerializedElement,
+} from '../press-contacts/lib';
```
kudlajz, line 12: Throughout this PR I am reusing a bunch of code from the `press-contacts` extension. I wasn't sure if it's worth it and where exactly to put this common functionality.

Assuming that we might possibly merge these two extensions into one in the future, is it okay like this (for now)?

packages/slate-editor/src/modules/editor/types.ts
```diff
@@ -105,6 +106,7 @@ export interface EditorProps {
     withGalleries?: false | GalleriesExtensionConfiguration;
     withHeadings?: boolean;
     withImages?: false | ImageExtensionConfiguration;
+    withInlineContacts?: false | InlineContactsExtensionParameters;
```
kudlajz, line 109: Is this a good name or do you think it should be different?

packages/slate-editor/src/modules/editor/getEnabledExtensions.ts
```diff
@@ -146,6 +149,10 @@ export function* getEnabledExtensions({
         yield HeadingExtension();
     }
 
+    if (withInlineContacts) {
+        yield InlineContactsExtension(withInlineContacts);
```
kudlajz, line 153: Example usage in this PR: https://github.com/prezly/prezly/pull/12333

packages/slate-editor/src/modules/editor/getEnabledExtensions.ts
```diff
@@ -146,6 +149,10 @@ export function* getEnabledExtensions({
         yield HeadingExtension();
     }
 
+    if (withInlineContacts) {
```
e1himself, line 152: I suggest adding a hard error in case both `withPressContacts` and `withInlineContacts` extensions are enabled. Just to make sure we don't shoot ourselves in foot.

packages/slate-editor/src/extensions/inline-contacts/transforms/updateInlineContact.ts
```diff
@@ -0,0 +1,14 @@
+import type { ContactNode } from '@prezly/slate-types';
+import type { Editor } from 'slate';
+import { Transforms } from 'slate';
+
+export function updateInlineContact(
+    editor: Editor,
+    contact: ContactNode,
+    patch: Partial<Pick<ContactNode, 'contact'>>,
```
e1himself, line 8: Why `Partial<...>`? Is there a point of calling this without providing `contact`?

packages/slate-editor/src/extensions/inline-contacts/InlineContactsExtension.tsx
e1himself: Agree with reusing them. :+1: 

I'd only suggest avoiding deep imports and explicitly exporting these via the `press-contacts` modue index.ts file.

packages/slate-editor/src/modules/editor/types.ts
e1himself: > I am still not sure about the name as inline to me implies something that can be inlined in a text (like inline links).
> 
> Thoughts?

I share your concern. But I can't come with any name that would be better and still remain clear for us, developers. 

Other options I have:
- ad-hoc contacts?
- local contacts?
- contact cards?

But I'm not liking any of the above, more than "inline contacts".

:shrug:

packages/slate-editor/src/extensions/inline-contacts/InlineContactsExtension.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+import { InlineContactElement } from './components';
+import type { InlineContactsExtensionParameters } from './types';
+
+export const EXTENSION_ID = 'InlineContactExtension';
+
+export function InlineContactsExtension({
+    renderForm,
```
e1himself, line 20: I'm questioning myself, what's the value of extracting the form rendering logic to the external code?

Left the same comment in https://github.com/prezly/prezly/pull/12333 too.

e1himself (commented):
Great job Lukas!

Please see my comments below.

packages/slate-editor/src/modules/editor/types.ts
kudlajz: I like `contact-cards` but it's not very clear for us, since we have two variations 🤔

packages/slate-editor/src/extensions/inline-contacts/transforms/updateInlineContact.ts
kudlajz: That's an oversight 👍 Initially I wanted to use `patch: Partial<ContactInfo>` but then changed it to `ContactNode` so it's consistent with other transformations. Will fix.

packages/slate-editor/src/modules/editor/types.ts
e1himself: > but it's not very clear for us, since we have two variations

:100: Exactly my thoughts.

packages/slate-editor/src/extensions/inline-contacts/InlineContactsExtension.tsx
kudlajz: One thing I wasn't sure about is where to put the form since it will be used in both the Placeholder extension as well as InlineContacts extension.

Only place I can think of is `src/components` but I am not sure as the form is not a generic component 🤔

kudlajz: Hmm, maybe `src/modules/components` could work?

packages/slate-editor/src/components/Input/Input.tsx
```diff
@@ -75,7 +75,9 @@ export const Input = forwardRef<HTMLInputElement, Props>(
         }, [value, pattern]);
 
         useEffect(() => {
-            setTimeout(() => input.current?.focus(), 0);
+            if (autoFocus) {
+                setTimeout(() => input.current?.focus(), 0);
+            }
```
kudlajz, line 80: I think this might have been a bug as it was always focusing the last rendered input regardless of `autoFocus` value.

I think the reason it was never apparent was because we only render one input at a time.

packages/slate-editor/src/components/Input/Input.tsx
```diff
@@ -110,6 +112,7 @@ export const Input = forwardRef<HTMLInputElement, Props>(
                             setFocused(false);
                             onBlur?.(event);
                         }}
+                        required={required}
```
kudlajz, line 115: I had to pass the prop here since it wasn't part of `attributes` to allow having validation on a form. I've also removed `required` from `SearchInputPlaceholder` in order to not render the error state if it's empty. Also doesn't seem like it needs to be required since the value is not used as far as I'm aware 🤔

packages/slate-editor/src/modules/components/InlineContactForm/InlineContactForm.tsx
```diff
@@ -0,0 +1,192 @@
...
+    }
+
+    useEffect(() => {
+        if (contact) {
+            setFormState(contact);
+        }
+    }, [contact]);
+
+    return (
+        <form onSubmit={handleSubmit}>
```
kudlajz, line 58: Screenshot 2023-01-20 at 16 26 07

The form currently looks like this. I did not add the icons there as we don't have any of those icons (phone, mobile, facebook, twitter) in this repository at the moment.

Do you think it's really important and I should add them or is it fine to skip it? I am a bit lazy here 😄

packages/slate-editor/src/components/Input/Input.tsx
kudlajz: One side effect of this is that the validation is immediately triggered, see:

Screenshot 2023-01-20 at 17 07 25

Maybe we can change it a bit so it only renders the validation state if the input becomes dirty? 🤔

e1himself: > One side effect of this is that the validation is immediately triggered, see:

So maybe the input shouldn't be marked as `required`? AFAIK the button will remain disabled, if it's empty. Right?

e1himself (approved):
Amazing work Lukas! :heart_eyes:

packages/slate-editor/src/components/Input/Input.tsx
kudlajz: I am not sure it's gonna be enough. According to [this code](https://github.com/prezly/slate/blob/main/packages/slate-editor/src/components/Input/Input.tsx#L132) it disables the button either when the value is invalid or if required but value is empty.

I think the `required && isEmpty` part is redundant, because empty value on a required input will be caught by the validation.

So, either we can keep it like this, add the dirty check or introduce a prop for disabling the button explicitly.

kudlajz: I went with the "dirty" check. Whenever `handleChange` is called, it will set the input as dirty and display the validation afterwards. So no immediate validation state when presented with empty input. Also dropped the extra condition. See https://github.com/prezly/slate/pull/372/commits/d90acface9960eea306ba835aa653308778f8ffa.

Let's see how it works and we can tweak it if necessary.

PR #377, opened 2023-01-24
[DEV-9877] Implement smart content-aware comparisons for other block types

e1himself:
Related to #375 

- `AttachmentNode`
- `BookmarkNode`
- `CoverageNode`
- `EmbedNode`
- `ImageNode`
- `GalleryNode`
- `StoryBookmarkNode`
- `StoryEmbedNode`
- `VideoNode`

### DIFF
--- a/packages/slate-editor/src/extensions/web-bookmark/WebBookmarkExtension.tsx
+++ b/packages/slate-editor/src/extensions/web-bookmark/WebBookmarkExtension.tsx
@@ -1,6 +1,7 @@
 import type { Extension } from '@prezly/slate-commons';
 import { createDeserializeElement } from '@prezly/slate-commons';
 import { BOOKMARK_NODE_TYPE, isBookmarkNode } from '@prezly/slate-types';
+import { isEqual } from 'lodash-es';
 import React from 'react';
 import type { RenderElementProps } from 'slate-react';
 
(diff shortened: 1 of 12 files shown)


packages/slate-editor/src/extensions/web-bookmark/WebBookmarkExtension.tsx
```diff
@@ -24,6 +25,19 @@ export const WebBookmarkExtension = ({
             [BOOKMARK_NODE_TYPE]: createDeserializeElement(parseSerializedElement),
         }),
     },
+    isElementEqual: (node, another) => {
+        if (isBookmarkNode(node) && isBookmarkNode(another)) {
+            // Compare ignoring `uuid` and `children`
+            return (
+                node.url === another.url &&
```
aspirisen, line 32: Am I right that we should not forget to update `isElementEqual` once we will add some new attributes to a node?

e1himself: Yes, you are right.

For this reason, I've been thinking about using `isEqual(omit())` instead of explicitly comparing every prop, but in the end decided that runtime performance gain (even a small one) is worth pursuing.

e1himself: Modification detection will start to produce false-negative results, marking the content as not modified, even though some property is different.

aspirisen (approved):

PR #378, opened 2023-01-24
[DEV-9878] ContactNode URLs get normalized when serialized/deserialized

pestaa:


packages/slate-editor/src/modules/components/InlineContactForm/InlineContactForm.tsx
```diff
@@ -4,10 +4,12 @@ import type { FormEvent } from 'react';
 import React from 'react';
 import { useEffect, useState } from 'react';
 
-import { Avatar, Button, Input } from '../../../components';
+import { Avatar, Button, Input } from '#components';
+import {normalizeHref} from "#lib";
```
kudlajz, line 8: There should be spaces around the import, re-run the lint.

e1himself (commented):
I did not realize that the problem is on the editor side. Sorry.

My suggestion: https://github.com/prezly/slate/pull/379

PR #382, opened 2023-01-29
[DEV-9918][DEV-10719] Switch to PNPM

aspirisen:
- Switch to `pnpm`
  - Force `resolution-mode=lowest-direct` (see https://github.com/pnpm/pnpm/issues/6463#issuecomment-1536115297)
- Bump Typescript language version to `4.9`
- Switch tests from `jest` to `mocha`
- Extract `storybook` dependencies into own sub-package


packages/slate-editor/jest.config.cjs
```diff
@@ -1,13 +1,21 @@
...
     extensionsToTreatAsEsm: ['.ts', '.tsx'],
     globals: {
         'ts-jest': {
             useESM: true,
+            isolatedModules: true,
             tsconfig: `${__dirname}/tsconfig.jest.json`,
         },
     },
     moduleNameMapper: {
+        '@prezly/slate-tables': path.join(__dirname, '../slate-tables/src'),
```
aspirisen, line 14: There is a problem with `jsdom` - it doesn't resolve modules like nodejs. Do not know why

e1himself (commented):
@aspirisen Do you have these warnings when you run `pnpm build`?

image

packages/slate-commons/src/plugins/withInlineVoid.ts
```diff
@@ -3,8 +3,10 @@ import type { Editor } from 'slate';
 
 import type { Extension } from '../types';
 
-export function withInlineVoid(getExtensions: () => Extension[]) {
-    return function <T extends Editor>(editor: T) {
+export function withInlineVoid(
+    getExtensions: () => Extension[],
+): <T extends Editor>(editor: T) => T {
+    return function <T extends Editor>(editor: T): T {
```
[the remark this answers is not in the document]
aspirisen: Fixed

aspirisen:
> Do you have these warnings when you run pnpm build?

Yes, switched to old lock file format

e1himself:
Works well now! Great! :+1: 

Also, let's squash all these commits (too many commits with crappy descriptions

).

aspirisen:
> Also, let's squash all these commits upon merge (too many commits with crappy descriptions).

I will merge with squash

packages/slate-editor/src/extensions/placeholders/PlaceholdersExtension.tsx
```diff
@@ -80,7 +80,6 @@ export function PlaceholdersExtension({
             return undefined;
         },
         isRichBlock: PlaceholderNode.isPlaceholderNode,
-        isVoid: PlaceholderNode.isPlaceholderNode,
```
aspirisen, line 83: @e1himself this is the reason why pasting didn't work - looks like slate detects isVoid element to disallow pasting. Placeholders are not void, they have inputs. What do you think?

e1himself: @aspirisen this doesn't sound right. 

I believe void vs not void is telling Slate how element `children` are handled. 

https://docs.slatejs.org/api/nodes/element#void-vs-not-void

> ### Void vs Not Void
> 
> In a not "void" element, Slate handles the rendering of its children (e.g. in a paragraph where the Text and Inline children are rendered by Slate). In a "void" element, the children are rendered by the Element's render code.

With this definition, Placeholder is a void node. As its `children` array doesn't mean anything and can only be empty (i.e. `[{ text: "" }]`).

aspirisen: hm, but the placeholder allows user input, we just render it as an input element, but we could render it as a `contenteditable` div that looks like the input and ux would be the same

I thought about void elements like about elements that does not imply user input, like images, dividers

kudlajz: But you would lose all the capabilities of `input` element, no? Like validation, pattern, focus etc.

aspirisen: I think it is implementation details, placeholders are part of slate ui and support text input

@e1himself what will happen if placeholders will be non-void elements?

aspirisen: This is tricky one, one the one hand we mark placeholder as an void element, but on the other slate for some reason forbids pasting into void elements

e1himself: > hm, but the placeholder allows user input, we just render it as an input element, but we could render it as a contenteditable div that looks like the input and ux would be the same

No, please. This is clearly not the right path.

> but on the other slate for some reason forbids pasting into void elements

This doesn't fit into my understanding of how Slate works.

Let's try to find the exact line of code that forbids it? Maybe this is a bug and we'll be able to report and fix it. Or maybe it's something else.

e1himself: > hm, but the placeholder allows user input, we just render it as an input element, but we could render it as a contenteditable div that looks like the input and ux would be the same

No, please. This is clearly not the right path.

> but on the other slate for some reason forbids pasting into void elements

This doesn't fit into my understanding of how Slate works.

Let's try to find the exact line of code that forbids it? Maybe this is a bug and we'll be able to report and fix it. Or maybe it's something else on our side.

aspirisen: Created an issue in slate https://github.com/ianstormtaylor/slate/issues/5294

e1himself:
@aspirisen My apologies for the delay. I'll review it today and merge. Thank you :pray:

packages/slate-editor/gulpfile.js
```diff
@@ -61,7 +63,19 @@ function buildEsm(files = JS_DELIVERABLE_SOURCES) {
...
         )
-        .pipe(babel({ extends: './babel.config.json' }))
+        .pipe(
+            babel({
+                extends: './babel.config.json',
+                plugins: [
+                    [
+                        'transform-define',
+                        {
+                            'process.env.ENV_SLATE_VERSION': packageJson.version,
```
aspirisen, line 73: @e1himself I just noticed that this is no accurate. We build slate before release and then this version will be outdated

e1himself (approved):

PR #385, opened 2023-02-02
[QRF-914] Update uploadcare

aspirisen:


packages/slate-editor/package.json
```diff
@@ -59,7 +59,7 @@
         "@prezly/slate-lists": "^0.76.3",
         "@prezly/slate-tables": "^0.76.3",
         "@prezly/slate-types": "^0.76.3",
-        "@prezly/uploadcare": "^2.4.0",
+        "@prezly/uploadcare": "^2.4.1",
```
e1himself, line 62: Hm... 

Is it necessary to ship a release like this? I mean, ``^2.4.0` already covers `^2.4.1`, so we bump the version in the app, it should also make the slate code using the newer version. Right?

aspirisen: good, will try to do that

PR #390, opened 2023-02-09
[QRF-918] Menu tethering

aspirisen:


packages/slate-editor/src/modules/popper-options-context/types.ts
```diff
@@ -5,7 +5,7 @@ export interface Context {
     modifiers?: {
         preventOverflow?: PreventOverflowModifier['options'];
     };
-    placement?: 'left-start' | 'right-start';
+    placement?: 'left' | 'right';
```
aspirisen, line 8: That was the original issue, allowing only `*-start` makes popper element to jump from top to bottom. Example: https://codesandbox.io/s/thirsty-taussig-8r8rzo?file=/src/index.js

packages/slate-editor/src/components/EditorBlock/getMenuPopperModifiers.ts
```diff
@@ -0,0 +1,115 @@
...
+            options: {
+                offset: ({
+                    popper,
+                    reference,
+                }: {
+                    popper: Rect;
+                    reference: Rect;
+                }): [number, number] => {
+                    const referenceHeight = getSlateElementHeight(reference.height);
+                    const popperTallerThanReference = popper.height - referenceHeight;
```
oleg-semyonov, line 38: I think it's better to call this variable `popperHeightDifference`. Current naming implies that this is a boolean.

packages/slate-editor/src/components/EditorBlock/getMenuPopperModifiers.ts
```diff
@@ -0,0 +1,115 @@
...
+                },
+            },
+        },
+        {
+            name: 'flip',
+            enabled: Boolean(popperOptions.autoPlacement),
+            options: {
+                // The order of these properties is important! The first one that has enough space to fit the popup will be used as fallback
+                // We prioritize flipping on Y axis (as this is the most common reason for overflow), then flipping X axis if needed.
+                fallbackPlacements: ['right', 'left'],
```
oleg-semyonov, line 62: Do I understand correctly, that `flip` modifier now only controls the horizontal overflow flip and the vertical flip is controlled exclusively by your new tether logic?
Asking because this replaces my fix in #359 that I made for https://github.com/prezly/prezly/pull/12099, so I might need to re-visit that.

packages/slate-editor/src/components/EditorBlock/Menu.tsx
```diff
@@ -103,9 +33,9 @@ export function Menu({ children, className, onClick, popperOptions, reference }:
     return (
         <Popper
             referenceElement={reference}
-            modifiers={getModifiers(popperOptions)}
-            placement={placement}
+            modifiers={getMenuPopperModifiers(popperOptions)}
             strategy="fixed"
+            placement={placement}
```
oleg-semyonov, line 38: ESLint should've complained about the prop order on this one 🤔

oleg-semyonov (approved):
Looks good! I've got one question and a couple remarks, outlined them in the comments.

packages/slate-editor/src/components/EditorBlock/getMenuPopperModifiers.ts
aspirisen: yeah, i tried to also some logic inside the name, just to indicate what i expect from this variable

packages/slate-editor/src/components/EditorBlock/getMenuPopperModifiers.ts
aspirisen: Because the menu now can stick to the edge of viewport i think previous behavior should work fine

packages/slate-editor/src/components/EditorBlock/Menu.tsx
aspirisen: Looks like we do not have such rule enabled here, i will do that in separate PR since there will be a lot of changes

oleg-semyonov: Then nevermind, I though we have the same rules enabled here

kudlajz (approved):
Looks great 👍 The only thing that confuses me is why did you decide to also update ESLint in the same PR? It probably should have been its own PR.

aspirisen:
@kudlajz ah right, forgot about that, i wanted to use `satisfies` to type popper modifiers, fixed here https://github.com/prezly/slate/pull/390/commits/17ce338e7018bcd5ae51d52b28e1e1413e757a90

PR #392, opened 2023-02-14
[CARE-729] Automatically wrap selection in link if pasted content is a valid URL

kudlajz:
link_pasting


aspirisen:
We also allow links in Quotes, Headings, and Tables. I think link pasting should also work there

oleg-semyonov (approved):
Nice!

kudlajz:
Good catch @aspirisen 👍 Addressed with https://github.com/prezly/slate/pull/392/commits/102292fda6165040b089b8e1003aee325694a463. Pasting in tables works already since it's using paragraphs inside table cells 🙂

PR #400, opened 2023-02-22
[CARE-1070] Do not provide progress for image origins that cannot be tracked

kudlajz:


packages/slate-editor/src/lib/fetchImageWithProgress/fetchImageWithProgress.ts
```diff
@@ -3,7 +3,7 @@ import { ProgressPromise } from '@prezly/progress-promise';
 import { fetchImageWithReadableStream } from './fetchImageWithReadableStream';
 import { fetchImageWithXmlHttpRequest } from './fetchImageWithXmlHttpRequest';
 
-const CORS_ENABLED_ORIGINS = ['https://cdn.uc.assets.prezly.com'];
+const CORS_ENABLED_ORIGINS = ['https://cdn.uc.assets.prezly.com', 'https://avatars-cdn.prezly.com'];
```
kudlajz, line 6: I am not sure if this is going to work for oembed screenshots, but don't have a way to test it locally for obvious reasons. Do you think this will work since the screenshots are loaded via restpack.io (the request to avatars-cdn is forwarded)? 🤔

e1himself: > Do you think this will work since the screenshots are loaded via restpack.io (the request to avatars-cdn is forwarded)?

Unlikely. The avatars endpoint _redirects_ the request to the real image location, not forwards.

kudlajz: Hmm 🤔

packages/slate-editor/src/lib/isSupportedImageOrigin.ts
```diff
@@ -0,0 +1,6 @@
+const CORS_ENABLED_ORIGINS = ['https://cdn.uc.assets.prezly.com'];
+
+export function isSupportedImageOrigin(url: string) {
```
kudlajz, line 3: Not sure if the name is okay or if it should be called something like `isCorsEnabledOrigin`? 🤔

packages/slate-editor/src/lib/hooks/useImage.ts
```diff
@@ -9,13 +10,21 @@ type OriginalURIOrBlobDataURI = string;
...
 export function useImage(src: string): State {
     const fetchImage = useCallback(() => fetchImageWithProgress(src), [src]);
     const { error, loading, progress, value } = useAsyncProgress(fetchImage);
-    return { error, loading, progress, loaded: Boolean(value), url: value };
+    return {
+        error,
+        loading,
+        // We can't track progress for unsupported origins and this way we can rely
+        // on the <ImageWithLoadingPlaceholder> component to estimate the progress.
+        progress: isSupportedImageOrigin(src) ? progress : undefined,
```
kudlajz, line 26: It's not possible to make the progress reporting work due to CORS so I am returning `undefined` here for unsupported origins so we get at least an approximation in the image component.

packages/slate-editor/src/lib/isSupportedImageOrigin.ts
oleg-semyonov: I think `isCorsEnabledOrigin` suits better, since the current name doesn't make clear what exactly is supported 🤔

oleg-semyonov (approved):

PR #405, opened 2023-04-25
[CARE-1379] Quickfix - Prezly editor is very slow

e1himself:
- Revert fix for CARE-1320 (see #404)
  The problem with this fix was that it's calling normalization _on absolutely every node_ in the document tree _on every keystroke_, which is A LOT for large documents. This was causing a major hit on the editor performance.
- Minor optimization on avoiding calling `Node.get()` in every normalizer

### DIFF
--- a/packages/slate-editor/src/modules/nodes-hierarchy/withNodesHierarchy.ts
+++ b/packages/slate-editor/src/modules/nodes-hierarchy/withNodesHierarchy.ts
@@ -18,7 +18,7 @@ export function withNodesHierarchy(schema: NodesHierarchySchema) {
             }
 
             for (const normalizer of normalizers) {
-                const isNormalized = normalizer(editor, path);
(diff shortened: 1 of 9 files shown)


packages/slate-editor/src/modules/nodes-hierarchy/withNodesHierarchy.ts
```diff
@@ -18,7 +18,7 @@ export function withNodesHierarchy(schema: NodesHierarchySchema) {
             }
 
             for (const normalizer of normalizers) {
-                const isNormalized = normalizer(editor, path);
+                const isNormalized = normalizer(editor, node, path);
```
e1himself, line 21: We have `node` here, so we can pass it to the normalizers. No need to call `Node.get()` to re-retrieve the node in every normalizer.

MohammadxAli (approved):

PR #412, opened 2023-05-11
[CARE-1320] Fix - Editor crashing when there is a node with `children: []`

e1himself:


packages/slate-commons/src/commands/insertNodes.ts
```diff
@@ -9,6 +9,7 @@ import { isBlock } from './isBlock';
 import { isCursorInEmptyParagraph } from './isCursorInEmptyParagraph';
 import { isInline } from './isInline';
 import { isVoid } from './isVoid';
+import { roughlyNormalizeNodes } from './roughly-normalize';
 
 interface Options {
```
MohammadxAli, line 14: _Unrelated question regarding this fix:_

I was curious about the absence of a `pnpm-lock.yaml` file in the root of this repository. I understand that there may be valid reasons for this decision, and I'm simply interested in understanding the rationale behind it.

e1himself: https://github.com/sindresorhus/ama/issues/479#issuecomment-310661514

PR #414, opened 2023-05-16
[DEV-10754] Add support for input addon in ContactPlaceholderElement

kudlajz:
This will allow us to render anything in front of the input.

Screenshot 2023-05-16 at 14 10 25


packages/slate-editor/src/components/Input/Input.module.scss
```diff
@@ -52,10 +52,6 @@ $focus-outline: $spacing-half solid #c7d2fe;
         outline: none;
     }
 
-    .focused & {
-        border-color: transparent;
-    }
```
e1himself, line 57: Why?

PR #415, opened 2023-05-16
[CARE-1627] Fix - Unignore storybook stories path for resolution

MohammadxAli:


packages/slate-editor/gulpfile.js
```diff
@@ -25,12 +25,7 @@ const BASE_DIR = './src';
 const SCSS_SOURCES = 'src/**/*.scss';
 const SCSS_GLOBAL_STYLESHEETS = 'src/styles/**/*.scss';
 const SCSS_MODULES = ['src/**/*.module.scss'];
-const TYPESCRIPT_SOURCES = [
-    'src/**/*.{ts,tsx}',
-    '!**/hyperscript.ts',
-    '!**/*.test.*',
-    '!**/*.stories.tsx',
-];
```
e1himself, line 33: There is an unwanted side-effect of this change: `*.stories.*` files will be included into the package releases.

Is there a way to avoid this?

e1himself (approved):

packages/slate-editor/gulpfile.js
MohammadxAli: I guess the best way is to make the storybook look into our `src` folder rather than `build`:

https://github.com/prezly/slate/blob/b8adefe745e1ff9281a23020f77c48252bda19cd/packages/slate-editor/storybook/config/main.cjs#L2

But that would require reworking that config entirely, and honestly, I don't have much experience there.

I'd say let's merge this as-is but tackle it in a separate issue: CARE-1633

e1himself: How about excluding `*.stories.*` from builds in production mode, but keep them in `dev` builds?

```diff
Index: packages/slate-editor/gulpfile.js
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/packages/slate-editor/gulpfile.js b/packages/slate-editor/gulpfile.js
--- a/packages/slate-editor/gulpfile.js	(revision b8adefe745e1ff9281a23020f77c48252bda19cd)
+++ b/packages/slate-editor/gulpfile.js	(date 1684325501636)
@@ -28,6 +28,8 @@
 const TYPESCRIPT_SOURCES = ['src/**/*.{ts,tsx}', '!**/hyperscript.ts', '!**/*.test.*'];
 const SVG_ICONS = 'src/**/*.svg';
 
+const STORYBOOK_MODULES = '*.stories.*';
+
 /**
  * Files that will produce build/* JS deliverables.
  */
@@ -54,6 +56,11 @@
                 stream.pipe(filter(SCSS_MODULES)).pipe(processSass()).pipe(filter('*.ts')),
             ]),
         )
+        .pipe(
+            filterOutStorybook({
+                enabled: process.env.NODE_ENV === 'production',
+            }),
+        )
         .pipe(
             babel({
                 extends: '../../babel.config.json',
@@ -79,6 +86,11 @@
                 stream.pipe(filter(SCSS_MODULES)).pipe(processSass()).pipe(filter('*.ts')),
             ]),
         )
+        .pipe(
+            filterOutStorybook({
+                enabled: process.env.NODE_ENV === 'production',
+            }),
+        )
         .pipe(compile())
         .pipe(
             inlineTypescriptAliases({
@@ -178,6 +190,13 @@
     ]);
 }
 
+function filterOutStorybook({ enabled = true } = {}) {
+    if (!enabled) {
+        return branch.obj((stream) => [stream]);
+    }
+    return branch.obj((stream) => [stream.pipe(filter(`!${STORYBOOK_MODULES}`))]);
+}
+
 /**
  * @param {string|string[]} files
  * @param {(string)=>void} incremental
```

MohammadxAli: Great! Yes, this definitely is a better solution: https://github.com/prezly/slate/pull/415/commits/ede59c1cd0ddeccf812fa47d12527ff33a549dde

PR #416, opened 2023-05-17
[CARE-1580] Refactor - Replace react-bootstrap's `Dropdown` with headlessui's `Listbox` in `Menu.Dropdown`

MohammadxAli:
There are two other places that need to be replaced:

- [MentionsDropdown](https://github.com/prezly/slate/blob/950cd7787fae1ec1a36ad3eb461ff8dbf5c0ef79/packages/slate-editor/src/extensions/mentions/components/MentionsDropdown.tsx#LL6C3-L6C3)
- [Dropdown (floating-add-menu)](https://github.com/prezly/slate/blob/950cd7787fae1ec1a36ad3eb461ff8dbf5c0ef79/packages/slate-editor/src/extensions/floating-add-menu/components/Dropdown.tsx#LL6C3-L6C3)

Will replace them in their corresponding PRs.

---

It now supports a focus trap as well as keyboard navigation:

https://github.com/prezly/slate/assets/67554982/df1f0383-ffcb-4358-a348-68862b387ce3

### DIFF
--- a/packages/slate-editor/src/components/Menu/Dropdown.tsx
+++ b/packages/slate-editor/src/components/Menu/Dropdown.tsx
@@ -1,9 +1,9 @@
+import { Listbox } from '@headlessui/react';
+import type { ListboxProps } from '@headlessui/react';
 import classNames from 'classnames';
 import type { ComponentType, FunctionComponent } from 'react';
-import { useState } from 'react';
+import { Fragment } from 'react';
 import React from 'react';
-import type { DropdownProps } from 'react-bootstrap';
-import { Dropdown as BootstrapDropdown, MenuItem } from 'react-bootstrap';
 
 import styles from './Dropdown.module.scss';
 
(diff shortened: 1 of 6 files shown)


packages/slate-editor/src/components/Menu/Dropdown.tsx
```diff
@@ -44,30 +45,36 @@ export function Dropdown<Value extends string = string>({
...
-            {...props}
+        <Listbox
             className={classNames(styles.Dropdown, className)}
-            onSelect={handleSelect}
-            open={open}
-            onToggle={(isOpen, _, meta) => meta.source !== 'rootClose' && setOpen(isOpen)}
+            as="div"
+            {...props}
+            value={selected}
+            onChange={handleSelect}
```
e1himself, line 53: I've noticed at least two regressions (in Firefox):

1. There is unwanted "focused" state ring around menu elements when I click around
2. There is a horizontal scrollbar appearing in the formatting dropdown menu

[Screencast from 2023-05-17 15-38-51.webm](https://github.com/prezly/slate/assets/370680/2bcc0798-50d6-45f9-9672-9cabb801d5ab)

e1himself: > I was testing with the attached editor in Prezly

Yeah, that's one of the reasons I've created CARE-1580 -- some problems are not noticeable in the Prezly application. Like implicit dependency on bootstrap.css.

So testing it outside the app seems to be a good idea too :)

e1himself (approved):
Great job Ali! :muscle:

PR #418, opened 2023-05-18
[DEV-10804] Implement menu with options for PressContactElement

kudlajz:
DEV-10804

- implemented toolbox
- implemented avatar toggle
- implemented layout (card, signature) option

Screenshot 2023-05-22 at 10 51 08


packages/slate-editor/src/extensions/press-contacts/components/PressContactElement.tsx
```diff
@@ -1,32 +1,72 @@
...
             // We have to render children or Slate will fail when trying to find the node.
             renderAboveFrame={children}
-            renderMenu={renderMenu}
+            renderMenu={(props) => {
+                if (renderMenu) {
+                    return renderMenu(props);
+                }
+
+                return (
+                    <PressContactMenu
```
kudlajz, line 52: I need to add this link but the URL needs to be provided via plugin props. Just wasn't sure what would be a good name/structure for this type of thing? Any ideas?

I was thinking just going with something like "menuLinks: {url, label}[]" 🤔 

Screenshot 2023-05-22 at 10 55 20

e1himself: > I was thinking just going with something like `menuLinks: {url, label}[]`

IDK, maybe.

FWIW when we needed to generate external links in other places, we've been passing _generate_ functions as extension parameters: 

https://github.com/prezly/slate/blob/e8ba684a9faa4a6a5304856bac7681f3a75c9ba6/packages/slate-editor/src/extensions/story-bookmark/types.ts#L14-L16

packages/slate-editor/src/icons/ContactLayoutCard.svg
```diff
@@ -0,0 +1,7 @@
+<svg width="49" height="48" viewBox="0 0 49 48" fill="none" xmlns="http://www.w3.org/2000/svg">
```
e1himself, line 1: I believe the icons are wrong.

They should rely on `currentColor` and `fill` CSS properties for the active state visualization:

https://github.com/prezly/slate/blob/0373c7a75633a9dcf8a0681084f79a3e2de24537/packages/slate-editor/src/components/Menu/Menu.module.scss#L129-L132

These two icons seem to have active/inactive state baked into their SVG files.

image

kudlajz: Good catch. I've fixed it by modifying the SVGs to use the same color and opacity as the other similar options: a773d4f39006f7bfe214e3bcf6513a1b338794e2

That piece of CSS that you've linked is not applied in this case I think because it's a different component.

packages/slate-editor/src/extensions/press-contacts/components/PressContactElement.tsx
```diff
@@ -1,32 +1,72 @@
...
+
+    const handleToggleAvatar = useCallback(
+        (showAvatar: boolean) => updatePressContact(editor, element, { show_avatar: showAvatar }),
+        [editor, element],
+    );
+    const handleChangeLayout = useCallback(
+        (layout: ContactLayout) => updatePressContact(editor, element, { layout }),
+        [editor, element],
+    );
+    const handleRemove = useCallback(() => removePressContact(editor), [editor, element]);
```
e1himself, line 37: Is `element` dependency necessary here?

e1himself (approved):

packages/slate-editor/src/extensions/press-contacts/components/PressContactElement.tsx
kudlajz: No 🙂 330059c18adbe5564a654ac60a76ddb593412860

PR #419, opened 2023-05-23
Fix - Compare options first before comparing references in contacts

kudlajz:
This is a follow up to #418.

I've noticed changing the options in the contact's menu did not trigger a value change and then realised I forgot to update the equality checking function.

### DIFF
--- a/packages/slate-editor/src/extensions/press-contacts/PressContactsExtension.tsx
+++ b/packages/slate-editor/src/extensions/press-contacts/PressContactsExtension.tsx
@@ -24,10 +24,7 @@ export const PressContactsExtension = (): Extension => ({
     },
     isElementEqual(element, another) {
         if (isContactNode(element) && isContactNode(another)) {
-            if (
-                !isEqual(element.layout, another.layout) ||
-                !isEqual(element.show_avatar, another.show_avatar)
-            ) {
+            if (element.layout !== another.layout || element.show_avatar !== another.show_avatar) {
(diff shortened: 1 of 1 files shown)


packages/slate-editor/src/extensions/press-contacts/PressContactsExtension.tsx
```diff
@@ -24,6 +24,13 @@ export const PressContactsExtension = (): Extension => ({
     },
     isElementEqual(element, another) {
         if (isContactNode(element) && isContactNode(another)) {
+            if (
+                !isEqual(element.layout, another.layout) ||
+                !isEqual(element.show_avatar, another.show_avatar)
+            ) {
+                return false;
+            }
```
e1himself, line 32: Why use `isEqual` for primitive values?

```suggestion
            if (element.layout !== another.layout || element.show_avatar !== another.show_avatar) {
                return false;
            }
```

e1himself (approved):

packages/slate-editor/src/extensions/press-contacts/PressContactsExtension.tsx
kudlajz: Ah, because at first I only did one `isEqual` call with two objects containing both of these properties but it didn't look nice when formatted so I've split it into two calls but didn't realize at that point it's not needed 🙈 Changed it.

Thanks!

PR #423, opened 2023-05-30
[DEV-7542] Add optional "role" property to HeadingNode

kudlajz:


packages/slate-types/src/nodes/HeadingNode.ts
```diff
@@ -19,3 +25,18 @@ export function isHeadingNode(value: any, type?: string): boolean {
         (type === undefined || value.type === type)
     );
 }
+
+export function isTitleHeadingNode(value: any): value is HeadingNode<typeof HEADING_1_NODE_TYPE>;
```
kudlajz, line 29: I think for our current use case it's better to only allow `role: title` for heading 1 and `role: subtitle` for heading 2 but I wasn't sure if we should enforce it here or let any heading have any role. What do you think? 🤔

e1himself: > but I wasn't sure if we should enforce it here or let any heading have any role.

I believe we have to enforce this on the normalization level -- make sure `title` and `subtitle` roles are only applied to `h1` and `h2` respectively, and remove any other `role` value if it happened to be set on a node.

kudlajz: Right, I've changed the functions to only check whether the node is a heading with a specific role: a832d09c8b380406f5dc7846550db400eefecec6

If I understand it correctly, the normalization of `role` property will happen in the Story editor, right? Since it's up to the consumer to decide which headings can be title/subtitle? Which in our case is going to be the first two headings.

Meaning I will not do the normalization within the extension itself as someone might decide to have multiple subtitles for example.

e1himself: > Meaning I will not do the normalization within the extension itself as someone might decide to have multiple subtitles for example.

I'd do this in the main extension. What's the value of allowing to have multiple subtitles?

kudlajz: That was just an example, but a bad one. I am just thinking about how to write such normalization that only keeps the `role` property on the first and second heading node but removes it from the rest 🤔

kudlajz: I think it would be easier to do the normalization outside the editor (via `allowed-blocks` extension) because there we know we want the `role` to only be on headings that are the second and third blocks and the rest of headings has no role.

Similar to email editor - there we want all the headings to not have any roles.

But more importantly - I think it's a bit weird to normalize `role` attribute in the editor since there won't be any way to change this attribute, right? So how can we say - there should always be one title and one subtitle if it changes case by case?

e1himself (approved):

PR #426, opened 2023-05-31
[DEV-10911] Render placeholders for Title/Subtitle headings, disable other features

kudlajz:
- renders "Add a title"/"Add a subtitle" for Title and Subtitle headings that have no text
- disables the floating add button for those nodes
- disables variables for those nodes
- no need to disable `#` and `##` shortcuts as those only work on paragraphs

### DIFF
--- a/packages/slate-editor/src/extensions/heading/components/HeadingElement.tsx
+++ b/packages/slate-editor/src/extensions/heading/components/HeadingElement.tsx
@@ -1,8 +1,11 @@
+import { EditorCommands } from '@prezly/slate-commons';
 import type { HeadingNode } from '@prezly/slate-types';
+import { HeadingRole } from '@prezly/slate-types';
 import { HEADING_1_NODE_TYPE, HEADING_2_NODE_TYPE } from '@prezly/slate-types';
 import classNames from 'classnames';
 import type { HTMLAttributes } from 'react';
 import React, { forwardRef } from 'react';
+import { useSlateStatic } from 'slate-react';
 
 import styles from './HeadingElement.module.scss';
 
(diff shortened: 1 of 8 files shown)


packages/slate-editor/src/extensions/variables/useVariables.ts
```diff
@@ -16,11 +19,31 @@ function placeholderToOption(placeholder: Variable): Option<Variable> {
...
+            const nodes = Array.from(
+                Editor.nodes(editor, {
+                    at: range,
+                    match: (node) => isTitleHeadingNode(node) || isSubtitleHeadingNode(node),
+                }),
+            );
+            return nodes.length === 0;
+        },
+        [editor],
+    );
```
kudlajz, line 42: We do not have any mechanism to disallow mentions functionality as far as I'm aware so I had to came up with something like this.

For now, it disables mentions for title and subtitle but we should consider checking other nodes as well. For example, it's possible to use placeholders in image caption but the story is not able to render it.

packages/slate-editor/src/extensions/heading/components/HeadingElement.tsx
```diff
@@ -12,15 +15,48 @@ interface Props extends HTMLAttributes<HTMLHeadingElement> {
...
+        const isSubtitle = element.role === HeadingRole.SUBTITLE;
+
+        const finalClassNames = classNames(className, styles.HeadingElement, {
+            [styles.title]: isTitle,
+            [styles.subtitle]: isSubtitle,
+            [styles.headingOne]: !isTitle && isHeadingOne,
+            [styles.headingTwo]: !isSubtitle && isHeadingTwo,
+        });
+
+        if (isTitle && EditorCommands.isNodeEmpty(editor, element)) {
```
kudlajz, line 33: I've considered using Slate's `renderPlaceholder` but that only seems to work when the editor is completely empty.

This way, we always render the placeholder as long as the nodes are empty.

packages/slate-editor/src/extensions/heading/components/HeadingElement.tsx
```diff
@@ -12,15 +15,48 @@ interface Props extends HTMLAttributes<HTMLHeadingElement> {
...
+            return (
+                <Heading {...props} ref={ref} className={finalClassNames}>
+                    {children}
+                    <span contentEditable={false} className={styles.placeholder}>
+                        Add a title
```
kudlajz, line 38: We might want to want these customizable? 🤔

packages/slate-editor/src/extensions/mentions/useMentions.ts
```diff
@@ -25,16 +26,17 @@ export interface Mentions<V> {
 
 export function useMentions<V>({
     createMentionElement,
+    isEnabled = () => true,
```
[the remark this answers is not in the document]
e1himself: https://github.com/prezly/slate/pull/426/commits/9dbf6f9b27392d2189f9da49d4f644d8984eafde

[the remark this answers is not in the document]
e1himself: https://github.com/prezly/slate/pull/426/commits/9dbf6f9b27392d2189f9da49d4f644d8984eafde

PR #427, opened 2023-06-02
[DEV-10912][DEV-10907] Feature - `placeholder:media` node type

e1himself:


packages/slate-editor/src/extensions/placeholders/PlaceholdersManager.ts
```diff
@@ -38,6 +38,13 @@ interface Data {
         file: ImageNode['file'];
         caption: string;
     };
+    [Type.MEDIA]: {
+        media: {
+            type: 'image';
+            file: ImageNode['file'];
+            caption: string;
+        };
```
e1himself, line 46: This is structured to support other non-image media types in the future.

PR #440, opened 2023-06-14
[CARE-1802] Refactor - `slate-lists` plugin to not mutate editor instance

e1himself:
# :warning: BC Breaking changes

## Changed

- `withLists()` does now include `withListsReact()` -- no need to call `withListsReact()` separately
- `withLists()` no longer mutates the Editor instance -- thus there is no more `ListsEditor` interface extending Slate `Editor`
- `ListsEditor.*` namespace API hasn't changed, except `isListsEditor()` method is now `isListsSupported()` and is not a type guard anymore
- `ListsEditor` instance methods are now available in the `ListsEditor` namespace:
   - as proxy functions to `ListsSchema` 
     (e.g. `ListsEditor.isListNode(editor, node)`)
   - available via `ListsEditor.getListsSchema(editor).*` 
     (e.g. `Listseditor.getListsSchema(editor)?.isListNode(node)`)

## Added 

- `withListsSchema()` -- bind a `ListsSchema` object to an editor instance
- `withListsNormalization()` -- attach lists normalization to the editor instance (does mutate editor instance)
- `withListsReact()` -- unchanged

All three initializer functions above are now executed as part of the all-in-one `withLists()` initializer. You may, however, decide to only pick parts of it by manually invoking individual initializers from the above list instead of `withLists()`.

# Migration guide

- if you use `withLists()` to initialize the editor, remove any extra calls to `withListsReact()`, as it's already a part of 
- Replace `ListsEditor.isListsEditor()` usages with `ListsEditor.isListsEnabled()` 
- Drop any usages of the `ListsEditor` interface type and replace it with a generic `Editor` type
- Replace any `ListsEditor` _instance_ methods (e.g. `editor.isListNode(node)`) with their proxies in the `ListsEditor` namespace (e.g. `ListsEditor.isListNode(editor, node)`)

### DIFF
--- a/packages/slate-lists/src/registry.ts
+++ b/packages/slate-lists/src/registry.ts
@@ -1,0 +1,27 @@
+import type { Editor } from 'slate';
+
+import type { ListsSchema } from './types';
+
+const EDITOR_LISTS_SCHEMA = new WeakMap<Editor, ListsSchema>();
+
+export function register(editor: Editor, schema: ListsSchema): void {
+    EDITOR_LISTS_SCHEMA.set(editor, schema);
+}
+
+export function unregister(editor: Editor): void {
+    EDITOR_LISTS_SCHEMA.delete(editor);
+}
+
+export function has(editor: Editor) {
+    return EDITOR_LISTS_SCHEMA.has(editor);
+}
+
+export function get(editor: Editor): ListsSchema {
+    const schema = EDITOR_LISTS_SCHEMA.get(editor);
+    if (!schema) {
+        throw new Error(
+            'This editor instance does not have a ListsSchema associated. Make sure you initialize it with withLists() before using ListsEditor functionality.',
+        );
+    }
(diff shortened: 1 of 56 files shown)


packages/slate-lists/src/registry.ts
```diff
@@ -0,0 +1,27 @@
+import type { Editor } from 'slate';
+
+import type { ListsSchema } from './types';
+
+const EDITOR_LISTS_SCHEMA = new WeakMap<Editor, ListsSchema>();
```
e1himself, line 5: I've tried this as an experiment to achieve the Lists plugin functionality without mutating the `Editor` instance. This will become helpful when/if we refactor editor extensions to be linked dynamically (CARE-1802).

oleg-semyonov (approved):

kudlajz (approved):
Great work! 👏 

office

PR #443, opened 2023-06-19
[CARE-1728][CARE-1858] Fix suggestions dropdown positioning

e1himself:


packages/slate-editor/src/extensions/mentions/components/MentionsDropdown.tsx
```diff
@@ -37,36 +52,50 @@ export const MentionsDropdown = <V extends object>({
...
-
-            const rect = RangeFix.getBoundingClientRect(domRange);
-
-            if (!rect) {
-                return;
-            }
-
-            const { left, height, top } = rect;
-            suggestionsPanelRef.current.style.top = `${top + window.pageYOffset + height}px`;
-            suggestionsPanelRef.current.style.left = `${left + window.pageXOffset}px`;
```
e1himself, line 56: This implementation based on `position: fixed` has stopped working after we've introduced a container element with `translate: 0 0` style (used by the new main navigation sidebar). This changes the containing block of the fixed-positioned element to the one with `translate: 0 0`, making the positioning not work anymore.

See https://developer.mozilla.org/en-US/docs/Web/CSS/position#fixed_positioning

packages/slate-editor/src/extensions/mentions/components/MentionsDropdown.tsx
```diff
@@ -37,36 +52,50 @@ export const MentionsDropdown = <V extends object>({
...
+                if (!domRange) {
+                    return EMPTY_RECT;
+                }
+                const rect = RangeFix.getBoundingClientRect(domRange);
+                if (!rect) {
+                    return EMPTY_RECT;
+                }
+                const { left, right, top, bottom, width, height } = rect;
+                return {
+                    left,
```
MohammadxAli, line 71: Wouldn't a `...rest` spread be shorter and more suitable here?

MohammadxAli (approved):

kudlajz (approved):

packages/slate-editor/src/extensions/mentions/components/MentionsDropdown.tsx
e1himself: Yeah, good point. This is not obvious, but we cannot use spread here. I've added a comment about it: 78268b68a09b5fd88a2ea6a9fc9be01f47b3ae38

MohammadxAli: Oh interesting, thanks!

PR #445, opened 2023-06-21
[CARE-1853] Fix - Pasting into title subtitle

e1himself:
- [Tweak HTML-pasting behaviour](https://github.com/prezly/slate/commit/2c5cafbabb082a42d6462357480f1a70237c2e4b)

  To only inherit copied elements block formatting when there are 2+ elements, otherwise we only paste inner contents, preserving the target element block formatting. For example:

  - When pasted h1 into a blockquote -- the blockquote will have the h1 contents, but will remain a blockquote.
  - When pasted h1 + h2 + paragraph -- the pasted content will replace the blockquote and force the pasted elements block formatting (h1 + h2 + paragraph)

  This seems to be better matching what's expected intuitively.

- [Tweak](https://github.com/prezly/slate/commit/a62f77329ecaa56eb38bb77c5e949d449c5336a5) [insertNodes](https://github.com/prezly/slate/commit/a62f77329ecaa56eb38bb77c5e949d449c5336a5) [to remove any empty block upon pasting](https://github.com/prezly/slate/commit/a62f77329ecaa56eb38bb77c5e949d449c5336a5)

  Previously, it was only replacing (i.e. removing after insertion) the target element if it was an empty *paragraph*. Now it does so for any empty block element: heading, paragraph, blockquote, etc.

### DIFF
--- a/packages/slate-editor/src/modules/editor/plugins/withDeserializeHtml/withDeserializeHtml.ts
+++ b/packages/slate-editor/src/modules/editor/plugins/withDeserializeHtml/withDeserializeHtml.ts
@@ -4,6 +4,7 @@ import { cleanDocx } from '@prezly/docx-cleaner';
 import type { Extension } from '@prezly/slate-commons';
 import { EditorCommands } from '@prezly/slate-commons';
 import type { Editor } from 'slate';
+import { Element } from 'slate';
 
 import { createDataTransfer } from '#lib';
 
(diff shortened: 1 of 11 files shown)


packages/slate-editor/src/modules/editor/plugins/withDeserializeHtml/withDeserializeHtml.ts
```diff
@@ -61,7 +62,18 @@ export function withDeserializeHtml(
 
                 // If there are no "nodes" then there is no interesting "text/html" in clipboard.
                 // Pass through to default "insertData" so that "text/plain" is used if available.
-                if (nodes.length > 0) {
+                if (
```
MohammadxAli, line 65: This comment should be on top of L76 I guess?

MohammadxAli (commented):
Tests are failing 👀

packages/slate-editor/src/modules/editor/plugins/withDeserializeHtml/withDeserializeHtml.ts
e1himself: https://github.com/prezly/slate/pull/445/commits/9a53d507ce1d6d1942470d0df889add4f5ac6cac

MohammadxAli (approved):
Nice 🙂👍

kudlajz (approved):
Great work 👍

PR #447, opened 2023-06-27
[CARE-1916] Remove `new` and `testing` labels from editor blocks

oleg-semyonov:
CARE-1916

### DIFF
--- a/packages/slate-editor/src/extensions/floating-add-menu/components/Dropdown.tsx
+++ b/packages/slate-editor/src/extensions/floating-add-menu/components/Dropdown.tsx
@@ -196,7 +196,7 @@ export function Dropdown<Action>({
                                                 option.isBeta ? styles.warning : styles.primary,
                                             )}
                                         >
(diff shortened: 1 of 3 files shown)


packages/slate-editor/src/extensions/floating-add-menu/components/Dropdown.tsx
```diff
@@ -196,7 +196,7 @@ export function Dropdown<Action>({
                                                 option.isBeta ? styles.warning : styles.primary,
```
e1himself, line 196: Should we also tweak the style of the badge? Like make it [gray-ish instead of yellow](https://prezly.slack.com/archives/C08PUBNGY/p1687523209208189?thread_ts=1687429136.303879&cid=C08PUBNGY)?

oleg-semyonov: e974a046ce96f5c254921156ae573d407b64d4a3

e1himself (approved):

PR #450, opened 2023-07-04
[CARE-2017] Forbid translating editor area

e1himself:
See https://stackoverflow.com/questions/12238396/how-to-disable-google-translate-from-html-in-chrome

### DIFF
--- a/packages/slate-editor/src/modules/editable/EditableWithExtensions.tsx
+++ b/packages/slate-editor/src/modules/editable/EditableWithExtensions.tsx
@@ -7,6 +7,7 @@ import type {
     RenderElement,
     RenderLeaf,
 } from '@prezly/slate-commons';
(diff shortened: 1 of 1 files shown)


e1himself:
Not neeeding a review, but you might find this fix interesting.

packages/slate-editor/src/modules/editable/EditableWithExtensions.tsx
```diff
@@ -109,6 +111,8 @@ export function EditableWithExtensions({
     return (
         <Editable
             {...props}
+            className={classNames(className, 'notranslate')}
+            translate="no"
```
e1himself, line 115: https://stackoverflow.com/questions/12238396/how-to-disable-google-translate-from-html-in-chrome

PR #451, opened 2023-07-05
Refactor - Remove `canvas` dependency

MohammadxAli:
It is not used anywhere, and `pnpm clean` followed by `pnpm install` works as expected without it, so it shouldn't be a peer dependency of another package AFAIK.

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,7 +72,6 @@
         "babel-plugin-module-resolver": "^5.0.0",
         "babel-plugin-transform-remove-imports": "^1.7.0",
         "branch-pipe": "^1.0.1",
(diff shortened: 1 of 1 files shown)


e1himself (approved):
Hm... I can't remember why I've added it.

The commit (74347ae611324a2cd202893c41b104dcd5f54d39) is about switching from Jest to Mocha. So maybe it was necessary for tests *at that moment*. But I see them passing now. 

Should be good to go :+1:

e1himself:
Now I know why it was there :)

image

MohammadxAli:
Ohhhh! :') Sorry

PR #453, opened 2023-07-06
[CARE-1932] Enable SelectionDecoration only when linking text

e1himself:
Making it work all the time is troublesome, as it modifies the DOM structure in response to the native DOM selection events, causing broken references and out-of-bounds errors on the DOMRange objects.

See CARE-1932 for details.


packages/slate-editor/src/extensions/decorate-selection/components/SelectionHighlight.module.scss
```diff
@@ -1,8 +1,7 @@
 @import "styles/variables";
 
 .selection {
-    border-radius: 4px;
-    box-shadow: 0px 2px 0px 1px $editor-selection-bg, 0px -2px 0px 1px $editor-selection-bg;
```
e1himself, line 5: I've also reset the selection decoration styling to match the native browser selection style to avoid styling changes jump when the decoration kicks in.

packages/slate-editor/src/modules/editor/Editor.tsx
```diff
@@ -498,105 +501,117 @@ export const Editor = forwardRef<EditorRef, EditorProps>((props, forwardedRef) =
...
-                            renderInput={() =>
-                                withSnippets.renderInput({
-                                    onCreate: submitFloatingSnippetInput,
-                                })
-                            }
-                        />
-                    )}
+                    <DecorationsProvider decorate={decorate}>
+                        {(combinedDecorate) => (
+                            <>
```
e1himself, line 506: Just added the wrapper.

packages/slate-commons/src/types/DecorateFactory.ts
```diff
@@ -1,9 +1,5 @@
 import type { Editor } from 'slate';
 
 import type { Decorate } from './Decorate';
-import type { Extension } from './Extension';
 
-export type DecorateFactory = <E extends Editor, X extends Extension>(
-    editor: E,
-    extension: X,
-) => Decorate;
```
e1himself, line 9: We never relied on the `extension` being passed to these. So I've dropped the parameter to simplify the code and reuse the interface in `useDecorationFactory()`, which wouldn't be possible otherwise.

packages/slate-editor/src/extensions/decorate-selection/decorateSelection.ts
```diff
@@ -0,0 +1,22 @@
...
+        const intersection = Range.intersection(editor.selection, {
+            anchor: { path, offset: 0 },
+            focus: { path, offset: node.text.length },
+        });
+        if (intersection) {
+            return [{ ...intersection, [SELECTION_MARK]: true }];
+        }
+    }
+    return [];
+}
```
e1himself, line 18: Moved as is without changes.

packages/slate-editor/src/modules/editor/Editor.tsx
```diff
@@ -187,6 +188,10 @@ export const Editor = forwardRef<EditorRef, EditorProps>((props, forwardedRef) =
     }, [autoFocus, editor]);
 
     useEffect(() => {
+        if (!blurOnOutsideClick || Math.random() >= 0) {
```
MohammadxAli, line 191: Isn't `Math.random()` always `>=` than 0? I don't get this :)

MohammadxAli (approved):
👍

packages/slate-editor/src/modules/editor/Editor.tsx
oleg-semyonov: Yeah, this looks a bit weird to me as well 🤔 
If it's intentional, a comment explaining this magic would be nice :D

packages/slate-editor/src/modules/decorations/context.ts
```diff
@@ -0,0 +1,15 @@
...
+import { noop } from '@technically/lodash';
+import { createContext } from 'react';
+
+type Detach = () => void;
+
+export type DecorationsConnector = {
+    attach(decorate: Decorate): Detach;
+};
+
+export const decorationsContext = createContext<DecorationsConnector>({
```
oleg-semyonov, line 11: I believe the naming convention for context objects is PascalCase, so that it doesn't look weird in JSX 🤔

packages/slate-editor/src/modules/decorations/useDecoration.ts
```diff
@@ -0,0 +1,16 @@
+import type { Decorate } from '@prezly/slate-commons';
+import { noop } from '@technically/lodash';
+import { useContext, useEffect } from 'react';
+
+import { decorationsContext } from './context';
+
+export function useDecoration(decorate: Decorate | undefined): void {
+    const { attach } = useContext(decorationsContext);
+
+    return useEffect(() => {
```
oleg-semyonov, line 10: This `return` seems redundant, since there is nothing to really return

packages/slate-editor/src/modules/editor/Editor.tsx
e1himself: Ouch! This was my debug code I forgot to remove. :see_no_evil: Thanks guys! 

I often use `Math.random() < 0` or `Math.random() >= 0` to disable a branch in code. As doing `|| false` or `|| true` causes TS compilation errors :)

e1himself: 56852fe352041753230e071214a0a075dc6b3bd1

packages/slate-editor/src/modules/decorations/useDecoration.ts
e1himself: 56852fe352041753230e071214a0a075dc6b3bd1

packages/slate-editor/src/modules/decorations/context.ts
e1himself: 56852fe352041753230e071214a0a075dc6b3bd1

PR #454, opened 2023-07-07
[CARE-1965] Fix - Infinite list structure normalization loop

e1himself:
The problem was in `normalizeOrphanListItem` / `normalizeOrphanListItemText` -- one was applying a fix, and the other one was reverting it back.

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,6 +72,7 @@
         "babel-plugin-module-resolver": "^5.0.0",
         "babel-plugin-transform-remove-imports": "^1.7.0",
         "branch-pipe": "^1.0.1",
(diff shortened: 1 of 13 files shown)


package.json
```diff
@@ -72,6 +72,7 @@
         "babel-plugin-module-resolver": "^5.0.0",
         "babel-plugin-transform-remove-imports": "^1.7.0",
         "branch-pipe": "^1.0.1",
+        "canvas": "^2.11.2",
```
e1himself, line 75: Screenshot from 2023-07-07 14-39-40

MohammadxAli (approved):

PR #460, opened 2023-08-14
[DEV-11322] Warn users if CTA is unsupported in their current theme

fgyimah:


packages/slate-editor/src/extensions/button-block/components/ButtonBlockMenu.tsx
```diff
@@ -45,6 +46,11 @@ interface Props {
     onRemove: () => void;
     value: FormState;
     withNewTabOption: boolean;
+    infoText?: {
+        text: string;
+        actionLabel?: string;
+        action?: () => void;
+    };
```
e1himself, line 53: I have a bad feeling about this interface. I think it's too much tailored to the use-case to the point it's almost hard-coded.

How about changing it to something more generic:

```ts
info: Array<string | { text: string, href: string } | { text: string, onClick: () => void }>
```

And then pass `info` as:

```ts
info={[
    "Your current theme doesn't support these options. ", 
    { text: 'Go to themes', onClick: ... },
]} 
```

what do you think?

fgyimah: I agree with this, making the update

fgyimah: Updated

e1himself (approved):

PR #461, opened 2023-08-16
[DEV-11387] 0 value showing in Button block menu when selected theme supports CTA buttons

fgyimah:


packages/slate-editor/src/extensions/button-block/components/ButtonBlockMenu.tsx
```diff
@@ -152,9 +152,9 @@ export function ButtonMenu({ info, onUpdate, onClose, onRemove, value, withNewTa
                 Button settings
             </Toolbox.Header>
 
-            {info && info.length && (
+            {Boolean(info && info.length) && (
                 <Toolbox.Section>
-                    <InfoText className={styles.info}>{info.map(renderInfoText)}</InfoText>
+                    <InfoText className={styles.info}>{info?.map(renderInfoText)}</InfoText>
```
e1himself, line 157: Just to share. 

A slightly more elegant solution could be to have a default value for `info` during destructuring:

```ts
export function ButtonMenu({ info = [], ... }: Props) {
```

and then only check for `.length`:

```tsx
            {info.length > 0 && (
                <Toolbox.Section>
                    <InfoText className={styles.info}>{info?.map(renderInfoText)}</InfoText>
```

e1himself (approved):

PR #469, opened 2023-08-22
[DEV-11426] Pilot two separate embed types: Youtube, Instagram

MohammadxAli:


packages/slate-editor/src/extensions/placeholders/PlaceholdersExtension.tsx
```diff
@@ -325,6 +339,20 @@ export function PlaceholdersExtension({
...
+                    return (
+                        <VideoPlaceholderElement
+                            attributes={attributes}
+                            element={element}
+                            fetchOembed={withVideoPlaceholders.fetchOembed}
+                            format={format}
+                            removable={removable}
+                        >
+                            {children}
+                        </VideoPlaceholderElement>
```
MohammadxAli, line 352: I know that the content of these placeholders will change, for now I'm just using regular video and embed placeholder elements until design is ready.

MohammadxAli:
Converting to draft to implement the menu items based on design

packages/slate-editor/src/extensions/placeholders/PlaceholdersExtension.tsx
```diff
@@ -211,6 +211,20 @@ export function PlaceholdersExtension({
...
+                            attributes={attributes}
+                            element={element}
+                            fetchOembed={withEmbedPlaceholders.fetchOembed}
+                            format={format}
+                            removable={removable}
+                        >
+                            {children}
+                        </EmbedPlaceholderElement>
+                    );
+                }
```
e1himself, line 226: I feel like having these extra IFs here makes it too crowded.

How about moving the conditions into the specific placeholder components?
E.g. the Instagram placeholder support will be implemented in the `EmbedPlaceholderElement`. Youtube will sit in the `VideoPlaceholderElement`. And so on.

What would you say?

packages/slate-editor/src/extensions/placeholders/elements/EmbedPlaceholderElement.tsx
```diff
@@ -79,6 +79,29 @@ export function EmbedPlaceholderElement({
...
+                inputDescription="Insert an embed URL and hit Enter"
+                inputPattern={URL_WITH_OPTIONAL_PROTOCOL_REGEXP.source}
+                inputPlaceholder="media.giphy.com/GIF"
+                inputAction="Add embed"
+                onSubmit={handleSubmit}
+            >
+                {children}
+            </InputPlaceholderElement>
+        );
+    }
```
e1himself, line 103: A humble suggestion on how to avoid code duplication here:

```tsx
function render(override: { 
    title?: string, 
    description?: string, 
    placeholder?: string, 
    action?: string,
 } = {}) {
   const { 
       title = 'Embed',
       description = 'Insert an embed URL and hit Enter',
       placeholder = 'media.giphy.com/GIF',
       action = 'Add embed',
   } = override;
   return (
        <InputPlaceholderElement
            {...props}
            element={element}
            // Core
            format={format}
            icon={PlaceholderEmbed}
            title={Title}
            description={Description}
            // Input
            inputTitle={title}
            inputDescription={description}
            inputPattern={URL_WITH_OPTIONAL_PROTOCOL_REGEXP.source}
            inputPlaceholder={placeholder}
            inputAction={action}
            onSubmit={handleSubmit}
        >
            {children}
        </InputPlaceholderElement>
    );
}

if (element.provider === PlaceholderNode.Provider.INSTAGRAM) {
    return render({ 
        title: 'Instagram',
        description: 'Insert an Instagram URL and hit Enter',
        placeholder: 'https://www.instagram.com/p/Cr-y_XyOyL9/',
        action: 'Embed',
    });
}

return render();
```

packages/slate-editor/src/extensions/embed/EmbedExtension.tsx
```diff
@@ -17,9 +17,15 @@ interface Parameters extends EmbedExtensionConfiguration {
...
 
+export enum Provider {
+    INSTAGRAM = 'instagram',
+    YOUTUBE = 'youtube',
+}
+
 export interface EmbedExtensionConfiguration {
     fetchOembed: (url: OEmbedInfo['url']) => Promise<OEmbedInfo>;
     showAsScreenshot: boolean;
+    providers: `${Provider}`[];
```
e1himself, line 28: Hm... :thinking: 

Now, when I see this, I start to believe this is not the best place for the property. 

These are the smells:

1. `Provider` enum is a full copy of the `PlaceholderNode.Provider`
2. Some of the providers have nothing to do with Embeds and actually produce a `VideoNode` (e.g.: YouTube)
3. The extension itself is also almost completely unrelated to the functionality: no new code is being added to it.

That being said, I believe it would be better to move this setting to the `FloatingAddMenuExtension`, as this setting is mainly controlling the menu options. Right?

Also, let's define the `FloatingAddMenuExtension` config property using the `PlaceholderNode.Provider` enum (to avoid duplication).

What do you think @MohammadxAli ?

MohammadxAli: > 2. Some of the providers have nothing to do with Embeds and actually produce a VideoNode (e.g.: YouTube)

I also felt that it seemed like a code smell when I was implementing it. Looking into the code and I can't find a better place except for the `FloatingAddMenuExtenstion` (as you suggested) to define the `Provider`. But still, mixed feelings.

Let's start with this 👍 I guess as we build we might find a better place for it.

e1himself: > better place except for the FloatingAddMenuExtenstion (as you suggested) to define the Provider.

For the record:

I suggest keeping the `Provider` enum in the `PlaceholderNode` type definition (as is). And just import it in the `FloatingAddMenuExtension`.

e1himself: > better place except for the FloatingAddMenuExtenstion (as you suggested) to define the Provider.

For the record:

I suggest keeping the `Provider` enum in the `PlaceholderNode` type definition (as is). And just import it in the `FloatingAddMenuExtension`.

MohammadxAli: Maybe... Special embeds should be their own extension?

e1himself: What is a special embed?

MohammadxAli: So I have been thinking about this.

Special embeds depend at least on 2 other extensions:
1. `withEmbeds`
2. `withVideos`
3. `withPlaceholders` (optional)

In order to use them, one should have enabled at least `withVideos` and `withEmbeds` extensions to start with because we're supporting a range of possible providers in an array.

So I guess it makes sense to make the special embeds an extension of their own, something like `withSpecialEmbeds` which has 2 dependencies (videos and embeds).

What do you think?

MohammadxAli: > What is a special embed?

"Specific embeds" group as we call them in the UI as well. Instagram, Youtube and etc...

e1himself: They for sure do not need their own extension.

Another thing is that the chosen configuration interface to enable them from outside might be suboptimal. :thinking: 

> withSpecialEmbeds which has 2 dependencies (videos and embeds)

This argument isn't strong. As the menu extension does already depend on almost every other block-type extension. So I wouldn't worry about it.

-- 

How about changing `specialEmbeds?: Provider[]` to `withFloatingAddMenu: { withSpecificProviderOptions?: boolean }`?

and then the menu will generate these additional options depending on the flags:

```ts
if (withEmbeds && withSpecificProviderOptions) {
   yield { /* Instagram menu option */ };
}

if (withVideos && withSpecificProviderOptions) {
   yield { /* YouTube menu option */ };
}
```

I believe this way it should resolve your concern on the extensions inter-dependency. They logic will reflect it with the IFs.

e1himself (approved):

PR #471, opened 2023-08-23
[DEV-11492] Enhance YouTube placeholder component and its menu item

MohammadxAli:


packages/slate-editor/src/modules/editor/menuOptions.tsx
```diff
@@ -210,6 +210,8 @@ function* generateOptions(
     if (withVideos && withSpecificProviderOptions) {
         yield {
             action: MenuAction.ADD_YOUTUBE,
+            isNew: true,
+            keywords: ['video', 'yt'],
```
MohammadxAli, line 214: @alpower I have used these keywords to make the YouTube option show up when users are searching through the [+] menu, letting you know for suggesting any improvements or changes. It would be nice to also do this for any other dedicated option that we're gonna have.

Related comment: https://linear.app/prezly/issue/DEV-11484#comment-72fcc4c9

packages/slate-editor/src/extensions/placeholders/elements/VideoPlaceholderElement.tsx
```diff
@@ -158,29 +165,45 @@ export function VideoPlaceholderElement({
...
     return render();
 }
 
-function Title(props: { isDragOver: boolean; isLoading: boolean }) {
-    if (props.isLoading) {
+interface TitleProps {
+    isDragOver: boolean;
+    isLoading: boolean;
+    text?: string;
+}
```
e1himself, line 184: I usually avoid declaring multiple Props interfaces in one file. For the main component there is `Props`. For the rest I do inline definition (as it was before) to not pollute the file visually.

Not like this is important, just sharing my perspective on it :)

e1himself (approved):

PR #475, opened 2023-08-24
[DEV-11495] Feature - Implement "Placeholders" and dedicated "Menu Item" for X (Twitter)

MohammadxAli:
# Menu Item

image

# Placeholders

image
image


MohammadxAli:
Since options are too many, I will merge each 3-4 options into a single PR from now on to avoid noise and make reviewing easier.

cc @e1himself @alpower

packages/slate-editor/src/extensions/placeholders/elements/SocialPostPlaceholderElement.tsx
```diff
@@ -136,6 +136,17 @@ export function SocialPostPlaceholderElement({
         });
     }
 
+    if (element.provider === PlaceholderNode.Provider.X) {
+        return render({
+            inputTitle: 'X post',
+            inputDescription: 'Paste a social media link and hit Enter',
+            inputPlaceholder: 'https://www.x.com/post/',
```
e1himself, line 143: I believe they haven't yet migrated the URLs to the new domain. So I suggest to keep it as `twitter.com/post/` to not confuse people. 

@alpower agree?

e1himself (approved):

packages/slate-editor/src/extensions/placeholders/elements/SocialPostPlaceholderElement.tsx
MohammadxAli: Good call 👍

Used "twitter" instead: https://github.com/prezly/slate/pull/475/commits/0f7a3c9db2ed7c58577edec64e301eb1293636bf

PR #477, opened 2023-08-24
[DEV-11515][DEV-11499][DEV-11500][DEV-11516] Implement dedicated Placeholders and Menu Items for TikTok, Spotify and Facebook

MohammadxAli:
# Facebook (DEV-11500)

| Menu Item | Placeholder Prompt | Placeholder Input |
|--------|--------|--------|
| image | image | image | 

# TikTok (DEV-11515)
| Menu Item | Placeholder Prompt | Placeholder Input |
|--------|--------|--------|
| image | image | image | 

# Spotify (DEV-11499)
| Menu Item | Placeholder Prompt | Placeholder Input |
|--------|--------|--------|
| image  | image  | image |


packages/slate-editor/src/extensions/placeholders/elements/SocialPostPlaceholderElement.tsx
```diff
@@ -147,6 +147,16 @@ export function SocialPostPlaceholderElement({
         });
     }
 
+    if (element.provider === PlaceholderNode.Provider.FACEBOOK) {
+        return render({
+            inputTitle: 'Facebook post',
+            inputDescription: 'Paste a social media link and hit Enter',
+            inputPlaceholder: 'https://www.facebook.com/user/post/',
+            inputAction: 'Add link',
+            title: (props) => <Title {...props} text="Click to embed an Facebook post" />,
```
alpower, line 156:
```suggestion
            title: (props) => <Title {...props} text="Click to embed a Facebook post" />,
```

alpower (approved):
approved with one suggestion made to change.

packages/slate-editor/src/icons/Component-TikTok.svg
```diff
@@ -0,0 +1,6 @@
+<svg width="32" height="32" viewBox="0 0 32 32" fill="none" xmlns="http://www.w3.org/2000/svg">
+    <rect width="32" height="32" rx="4.8" fill="#374151" />
+    <path fill-rule="evenodd" clip-rule="evenodd"
```
MohammadxAli, line 3: _Comment in an unrelated place_

Hey @e1himself, there's a problem that I see with SVGs which I'm not sure how to solve.

So we're using an SVG optimizer that minifies the classes and IDs for SVG parts, this will result in a conflicting situation with some of the SVGs sharing the same id for their clip-path URL with the same ID, see here:

Giphy:
image

Vimeo:
image

Do you have any solution in mind?

alpower: "Some, not all" points to the original SVGs (and I can confirm that these have clip paths when the others don't). Let me try and make you some that don't have these clip paths.

An aside, but I think that maybe we put new on the 'embed' section menu, instead of on every embed? think on every one looks a bit overwhelming (I know you won't see this when filtered however?)

MohammadxAli: OK, a little more research:

1. SVGR is using SVGO under the hood.
2. SVGO is using [`cleanupIds` plugin](https://github.com/svg/svgo/blob/main/plugins/cleanupIds.js) to minify and cleanup the ids.
3. We need a way to tell SVGR to disable that plugin.

But I'm not sure how to do it in our build system. I am using the web GUI here: https://jakearchibald.github.io/svgomg/

And if you disable the "Clean up IDs" plugin, it won't try cleaning up the IDs.

MohammadxAli: >  (I know you won't see this when filtered however?)

That's true, because when filtered, there is only 1 SVG with the clip path ID of `#a` for example.

> Let me try and make you some that don't have these clip paths.

But this is a problem with our build system in its current form, we should be able to solve it correctly to avoid this stuff from happening in the future. However I can accept that as a middle ground solution :-)

alpower: @MohammadxAli unless your tool is automated, and you can't control the output, I'd change the reference ald link to reference to something unique i.e. something like this will make it unique compared to the other SVG.
CleanShot 2023-08-24 at 14 49 26@2x

MohammadxAli: Well, that's the problem, the SVGs I have added to the codebase already use a unique ID.

But our build system which uses SVGR/SVGO turns those unique IDs into non-unique IDs :')

`blah-blah-blah` -> `a`
`meh-meh-meh` -> `a`

BOOM.

MohammadxAli: Here is the complete list of SVGs that use clip-path which potentially will result in conflicting IDs:

- Component-GoogleDocs
- Component-GoogleSheets
- Component-Tally
- Component-Giphy
- Component-Vimeo

cc @alpower, can you somehow remove the clip-path from those in Figma so that I can re-export them?

alpower: Fixed the source Figma file so no clip paths exist. ✅

e1himself: Hey :wave: 

Good job figuring out how to work it around :trophy: 

> But I'm not sure how to do it in our build system. I am using the web GUI here: https://jakearchibald.github.io/svgomg/
> And if you disable the "Clean up IDs" plugin, it won't try cleaning up the IDs, which I expect to solve the issue.

Just for future reference, I believe there is a way to configure this:
- The Babel transformation plugin we use for SVGs:
  https://github.com/birdofpreyru/babel-preset-svgr#under-the-hood
- SVGR configuration reference:  
  https://react-svgr.com/docs/options/#svgo
- svgo.config.js reference:
  https://github.com/svg/svgo#configuration

But just to be clear: let's not spend time on fixing the configuration now :)

e1himself (approved):

PR #489, opened 2023-08-31
[DEV-11518] Feature - Render Tally embeds as web bookmarks in emails

e1himself:
The main change here is restructuring the Embed element rendering to fallback to the Bookmark presentation. This happens when both iframe HTML and Screenshot presentations are either disabled or not possible.

The rest of changes are refactoring to simplify the reuse of Bookmark presentation


MohammadxAli (approved):
👌

PR #494, opened 2023-09-01
[DEV-11548] Fix - Address incorrect bookmark to video/embed interactions

MohammadxAli:


e1himself (approved):

packages/slate-editor/src/extensions/web-bookmark/components/WebBookmarkMenu.tsx
```diff
@@ -44,16 +44,19 @@ const LAYOUT_OPTIONS: OptionsGroupOption<BookmarkNode.Layout>[] = [
...
+            value: 'embed',
+            label: 'Embed',
+            disabled: !element.oembed.html,
+        },
+        {
+            value: 'card',
+            label: 'Bookmark',
+        },
+    ] as OptionsGroupOption<Presentation>[];
+}
```
e1himself, line 59: I usually avoid `as` casting, when possible.

```suggestion
function getPresentationOptions(element: BookmarkNode): OptionsGroupOption<Presentation>[] {
    return [
        {
            value: 'embed',
            label: 'Embed',
            disabled: !element.oembed.html,
        },
        {
            value: 'card',
            label: 'Bookmark',
        },
    ];
}
```

packages/slate-editor/src/extensions/web-bookmark/transforms/convertWebBookmark.ts
```diff
@@ -13,6 +14,18 @@ export function convertWebBookmark(
...
+                editor,
+                createVideoBookmark({ oembed: element.oembed, url: element.url }),
+                {
+                    at: [],
+                    match: (node) => node === element,
+                },
+            );
+            return;
+        }
+
```
e1himself, line 28: It's better to branch for the construction of the replacement element. As the rest of `setNodes()` call is exactly the same.

```ts
const converted = element.oembed.type === 'video' 
    ? createVideoBookmark(...)
    : createEmbed(...)
```

PR #498, opened 2023-09-06
[DEV-11547] Feature - Implement switching Bookmark/Embed/Video to/from InlineLink

e1himself:
[Screencast from 2023-09-07 12-32-11.webm](https://github.com/prezly/slate/assets/370680/5b8ff5aa-1515-48db-a698-4aaeedb77155)


packages/slate-editor/src/modules/rich-formatting-menu/RichFormattingMenu.tsx
```diff
@@ -186,9 +194,11 @@ export function RichFormattingMenu({
                 <LinkMenu
                     node={link}
                     canUnlink={link !== null}
+                    withConversionOptions={withConversionOptions}
                     withNewTabOption={withNewTabOption}
                     onBlur={clearLinkRange}
                     onChange={linkSelection}
+                    onConvert={withConversionOptions ? onConvert : undefined}
```
oleg-semyonov, line 201: `onConvert` will already be `undefined` if `withConversionOptions` is not set (line 178), so this ternary seems redundant 🤔

packages/slate-commons/src/commands/replaceNode.tsx
```diff
@@ -1,27 +1,35 @@
...
+            if (isElementNode(newNode)) {
+                replaceChildren(editor, [node, path], newNode.children);
+            }
+            if (select) {
+                Transforms.select(editor, path);
+            }
+            // exit after the fist iteration, otherwise there's risk that we'll plant
+            // multiple instances of the same object into the document. Which is bad:
+            // Slate is relying on referential equality and uniqueness internally.
+            return;
```
oleg-semyonov, line 32: If we only need to iterate once, wouldn't it be cleaner to just get the first item from `Editor.nodes(...)`?

e1himself: Somehow it looks cleaner to me this way, as it naturally handles the zero results case without extra hassle  :thinking: 

```ts
for (const [node, path] of Editor.nodes(...)) {
    // ...
    return // ... ;
}
```

vs 

```ts
const [entry] = Editor.nodes(/* ... */);
if (entry) {
   const [node, path] = entry;
   // ...
}
```

Don't you think it's more elegant with the interrupted loop? Or am I absusing it? :)

oleg-semyonov: The zero results case makes sense, and you do save one line, bit I still thinks that's a bit hacky ;)

You can also return early `if (!entry)`, and then the code would be pretty elegant 🤔

e1himself: > You can also return early if (!entry), and then the code would be pretty elegant 🤔

That's debatable :) 

> [...] but I still thinks that's a bit hacky ;)

Ok, changing to an `if` then :+1:

packages/slate-editor/src/modules/rich-formatting-menu/RichFormattingMenu.tsx
e1himself: 631d1431a1d3703ab84dcfaf50069e5bf207ed0b

packages/slate-commons/src/commands/replaceNode.tsx
e1himself: 931f99d4192bc9dad2c124005bd34eac4d580060

PR #502, opened 2023-09-13
[CARE-3062] Fix editor not seeing modifications Video node `layout` prop

e1himself:


packages/slate-editor/src/extensions/video/VideoExtension.tsx
```diff
@@ -41,8 +41,8 @@ export function VideoExtension({
             if (VideoNode.isVideoNode(node) && VideoNode.isVideoNode(another)) {
                 return (
                     node.url === another.url &&
-                    isEqual(node.oembed, another.oembed) &&
-                    node.layout === node.layout
```
e1himself, line 45: This was the problem: it compared a value to itself.

```ts
node.layout === node.layout
```

PR #509, opened 2023-12-11
[CARE-3589] Drop React v16 and v17 support

e1himself:
I don't think we use older React versions anywhere. Right?


oleg-semyonov (approved):
👍🏻

PR #525, opened 2024-01-31
[CARE-4306] Update outdated dependencies with security issues

kudlajz:
Also updated the `lerna` package with the latest one and replaced oudated `ts-plugin-scss-modules` with `typescript-plugin-css-modules`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,7 @@
         "@babel/preset-react": "^7.23.3",
         "@babel/preset-typescript": "^7.23.3",
         "@babel/register": "^7.23.7",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -129,5 +129,10 @@
             "slate-hyperscript",
             "slate-react"
         ]
+    },
+    "pnpm": {
+        "overrides": {
+            "glob-parent@<5.1.2": ">=5.1.2"
```
kudlajz, line 135: The reason for this override is because of the `gulp` package, which doesn't have a newer version. So, until we replace the `gulp` build pipeline with something else, we have to use this override.

PR #527, opened 2024-02-20
[DEV-12522] Feature - Implement inline drag & drop, captions, crop and delete in galleries

kudlajz:
Required changes for https://github.com/prezly/prezly/pull/14475.

### DIFF
--- a/packages/slate-editor/src/components/EditorBlock/EditorBlock.tsx
+++ b/packages/slate-editor/src/components/EditorBlock/EditorBlock.tsx
@@ -204,7 +204,6 @@ export const EditorBlock = forwardRef<HTMLDivElement, Props>(function (
                         [styles.overflowHidden]: overflow === 'hidden',
                         [styles.overflowVisible]: overflow === 'visible',
                         [styles.selected]: isSelected,
-                        [styles.fullWidth]: layout === 'full-width',
                     })}
                     onClick={handleFrameClick}
                     onMouseEnter={() => setIsHovered(true)}
--- a/packages/slate-editor/src/components/EditorBlock/EditorBlock.tsx
+++ b/packages/slate-editor/src/components/EditorBlock/EditorBlock.tsx
@@ -2,7 +2,6 @@ import { EditorCommands } from '@prezly/slate-commons';
 import type { ElementNode } from '@prezly/slate-types';
 import { Alignment } from '@prezly/slate-types';
 import classNames from 'classnames';
-import { isHotkey } from 'is-hotkey';
 import type { MouseEvent, ReactNode } from 'react';
 import React, { forwardRef, useCallback, useEffect, useState } from 'react';
 import { Editor, Transforms } from 'slate';
@@ -135,17 +134,6 @@ export const EditorBlock = forwardRef<HTMLDivElement, Props>(function (
         },
         [isOnlyBlockSelected],
     );
-
-    useEffect(() => {
-        const onEsc = (e: KeyboardEvent) => {
-            if (isHotkey('esc', e)) {
-                closeMenu();
-            }
-        };
-
-        document.addEventListener('keydown', onEsc);
-        return () => document.removeEventListener('keydown', onEsc);
-    }, [closeMenu]);
 
     const renderProps: RenderProps = {
         isSelected,
--- a/packages/slate-editor/src/extensions/galleries/components/Gallery/Gallery.tsx
+++ b/packages/slate-editor/src/extensions/galleries/components/Gallery/Gallery.tsx
@@ -44,7 +44,7 @@ export function Gallery({
 }: Props) {
     const [activeId, setActiveId] = useState<UniqueIdentifier | null>(null);
     const [orderedImages, setOrderedImages] = useState(images);
-    const ids = useMemo(() => orderedImages.map((image) => image.uuid), []);
+    const ids = useMemo(() => orderedImages.map((image) => image.uuid), [orderedImages]);
     const activeIndex = activeId ? orderedImages.findIndex((image) => image.uuid === activeId) : -1;
 
     const sensors = useSensors(useSensor(PointerSensor));
@@ -83,7 +83,7 @@ export function Gallery({
 
     const getIndex = useCallback((id: UniqueIdentifier) => {
         return orderedImages.findIndex((image) => image.uuid === id);
-    }, []);
+    }, [orderedImages]);
 
     return (
         <DndContext
--- a/packages/slate-editor/src/extensions/galleries/components/Gallery/Gallery.tsx
+++ b/packages/slate-editor/src/extensions/galleries/components/Gallery/Gallery.tsx
@@ -9,11 +9,15 @@ import {
 } from '@dnd-kit/core';
 import { arrayMove, SortableContext } from '@dnd-kit/sortable';
 import type { GalleryImage, GalleryNode } from '@prezly/slate-types';
-import { UploadcareImage } from '@prezly/uploadcare';
+import { awaitUploads, UploadcareImage } from '@prezly/uploadcare';
 import { isUploadcareImageSizeValid } from '@prezly/uploadcare';
+import { noop } from '@technically/lodash';
 import classNames from 'classnames';
 import type { HTMLAttributes } from 'react';
 import React, { useCallback, useMemo, useState } from 'react';
+import { useSlateStatic } from 'slate-react';
+
+import { UploadcareEditor } from '#modules/uploadcare';
 
 import { IMAGE_PADDING, IMAGE_SIZE } from './constants';
 import styles from './Gallery.module.scss';
@@ -46,6 +50,7 @@ export function Gallery({
     width,
     ...attributes
 }: Props) {
+    const editor = useSlateStatic();
     const [activeId, setActiveId] = useState<UniqueIdentifier | null>(null);
     const ids = useMemo(() => images.map((image) => image.file.uuid), [images]);
     const activeIndex = activeId ? images.findIndex((image) => image.file.uuid === activeId) : -1;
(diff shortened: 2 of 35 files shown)


packages/slate-editor/src/modules/editor/getEnabledExtensions.ts
```diff
@@ -235,8 +235,8 @@ export function* getEnabledExtensions(parameters: Parameters): Generator<Extensi
     if (withGalleries) {
         yield GalleriesExtension({
             availableWidth,
-            onEdited(editor, gallery, { failedUploads }) {
-                EventsEditor.dispatchEvent(editor, 'gallery-edited', {
+            onAdded(editor, gallery, { failedUploads }) {
```
kudlajz, line 238: Since the behaviour has changed, "add" will now only append images to the end of the gallery instead of editing the whole gallery.

Do you think it's worth keeping the "onEdited" callback/event in addition to having the `onAdd/onAdded` callback?

`onEdited` would still receive the same data as before.

packages/slate-editor/src/components/EditorBlock/EditorBlock.tsx
```diff
@@ -136,17 +135,6 @@ export const EditorBlock = forwardRef<HTMLDivElement, Props>(function (
         [isOnlyBlockSelected],
     );
 
-    useEffect(() => {
-        const onEsc = (e: KeyboardEvent) => {
```
kudlajz, line 140: I've raised this to Al and he agreed to get rid of it, but I think having the ability to close the menu is not really useful since the block still stays selected afterwards.

So I think it's better to keep the menu open as long as the block is selected.

packages/slate-editor/src/extensions/image/components/ImageElement.tsx
```diff
@@ -88,6 +88,7 @@ export function ImageElement({
                 files: [initialFileInfo],
                 imagesOnly: true,
                 multiple: false,
+                tabs: [],
```
kudlajz, line 91: Previously, when you clicked "Crop" in the image menu, it would spawn Uploadcare with the default tabs, so you could either crop the image or select a completely different one, which I think is really confusing behaviour.

Checked with Al and he agreed, so I've added the `tabs` property here which makes Uploadcare only show the cropping tool.

packages/slate-editor/src/extensions/galleries/components/Gallery/Gallery.tsx
```diff
@@ -1,69 +1,274 @@
...
     size,
+    uuid,
     width,
     ...attributes
 }: Props) {
+    const editor = useSlateStatic();
+    const [activeId, setActiveId] = useState<UniqueIdentifier | null>(null);
+    const images = useMemo(
+        () =>
+            originalImages.map((image, index) => ({ ...image, id: `${image.file.uuid}-${index}` })),
```
kudlajz, line 57: Since we rely on Uploadcare's UUID for keys and drag & drop ids, it stops working correctly as soon as you upload the same image twice, creating a UUID conflict.

I've spoked about this with Al and we believe it is okay to remove duplicate images from the gallery, but I am not 100 % sure to pull the trigger. We could write a normalization for this.

So, for now, I've implemented this extra logic that generates a unique `id` for each image in the gallery. This way, we can still support duplicate images, but it makes the code a bit more complex.

A small downside of this approach is the fact that I'm using index to generate the id, so some re-rendering of gallery tiles will be unavoidable, since the index will change for each image in the array that has to change a position.

packages/slate-editor/src/extensions/galleries/components/Gallery/Gallery.tsx
```diff
@@ -1,69 +1,274 @@
...
+    }
+
+    const getIndex = useCallback(
+        (id: UniqueIdentifier) => {
+            return images.findIndex((image) => image.id === id);
+        },
+        [images],
+    );
+
+    let tileIndex = 0;
```
kudlajz, line 176: I'm using this variable to keep track of the current index in the `images` since `calculatedLayout` is using nested arrays (this was the simplest approach pretty much).

packages/slate-editor/src/extensions/galleries/components/Gallery/Gallery.tsx
e1himself: > so ideally I still think we should remove duplicate images so we can rely on just the Uploadcare's UUID

I don't think this is a good idea.

There can be different images with the same `uuid`. Imagine uploading a large photo with 3 people, but then cropping it to a different person every time. This way you'll have three different photos with exactly the same `uuid`.

--- 

I believe the best way to fix it would be to stop relying on `image.file.uuid` as an identifier and instead add a new  autogenerated unique`image.uuid` property. Not for now, of course.

packages/slate-editor/src/extensions/galleries/components/Gallery/GalleryTile.tsx
```diff
@@ -1,37 +1,93 @@
...
     return (
         <div
             className={classNames(styles.GalleryTile, className, {
                 [styles.withBorderRadius]: withBorderRadius,
+                [styles.interactive]: isInteractive,
+                [styles.dragging]: dragging,
+                [styles.insertBefore]: insertPosition === 'before',
+                [styles.insertAfter]: insertPosition === 'after',
+                [styles.withMediumPadding]: padding === GalleryPadding.M,
+                [styles.withLargePadding]: padding === GalleryPadding.L,
```
e1himself, line 63: I've noticed that if you drag quickly enough, the tiles first get the hover state and then immediately lose it. Not sure if it's possible to fix it.

[Screencast from 2024-02-23 19-48-42.webm](https://github.com/prezly/slate/assets/370680/18ffef8b-cfb3-4780-91cf-3d6aa3cf589f)

e1himself (commented):
Amazing job @kudlajz! :muscle: 

I've tried to poke some holes in it and found a few issues. Not sure if any of them are blockers, up to you and Al. Please see below.

packages/slate-editor/src/extensions/galleries/components/Gallery/SortableGalleryTile.tsx
```diff
@@ -0,0 +1,31 @@
...
+    ...props
+}: Props & { getIndex: (id: UniqueIdentifier) => number; id: UniqueIdentifier }) {
+    const { active, attributes, isDragging, listeners, over, setNodeRef } = useSortable({ id });
+
+    let insertPosition: 'before' | 'after' | undefined = undefined;
+    if (active && active.id !== id && over && over.id === id) {
+        const activeIndex = getIndex(active.id);
+        const index = getIndex(id);
+        insertPosition = activeIndex > index ? 'before' : 'after';
+    }
```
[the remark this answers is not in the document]
kudlajz: Yeah, I was hoping there would be a simpler way to detect whether the "collision" happened on the edge and then depending on whether it's a left or right edge, I would show the "indicator". I think it will require manually detecting the collision via the bounding box. I would have to take a look at that.

There's another problem with edges though and that is, if you have one image on the edge of the row and then another image that's on the next row on the opposite edge. Due to the way our data is structured and how the algorithm works, it's for example impossible in this scenario to drag the image from 2nd row to the end of the 1st row and expect anything to happen (since the two images are already "next to each other" in the array).

From the user point of view, I would expect the image from the 2nd row to get "squeezed" to the end of the 1st row but that's unfortunately not how our layout algorithm works as it has no concept of rows, as the data structure is just a list. So, I think even if were able to show the "indicator" at the edge of the row, there's no guarantee that the image would be placed there as it's ultimately up to the algorithm to decide.

So, I am not sure how this can be improved without completely reworking the layout algorithm, which I haven't done for obvious reasons 🙂 

Screenshot 2024-02-26 at 12 01 44

packages/slate-editor/src/extensions/galleries/components/Gallery/GalleryTile.tsx
```diff
@@ -1,37 +1,93 @@
...
+                    <div className={styles.Actions}>
+                        <CropButton isSmall={withSmallButtons} onCrop={onCrop} />
+                        <DeleteButton isSmall={withSmallButtons} onDelete={onDelete} />
+                    </div>
+                    <div className={styles.DragHandle} {...props} tabIndex={-1} />
+                    <div
+                        className={classNames(styles.Caption, {
+                            [styles.visible]: caption !== '',
+                        })}
+                    >
```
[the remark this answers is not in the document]
kudlajz: The problem here with the outline is that on full width blocks, the outline is rendered as "inner" instead of the regular one.

I've tried removing it and keeping the regular outline and that fixes it. Obviously, there's now the problem that the outline overflows the container a little bit but I think it's fine in this case as otherwise I don't see a way to fix this as the inner outline takes away some space from the block.

cc @alpower 

Screenshot 2024-02-26 at 12 34 51

packages/slate-editor/src/extensions/galleries/components/Gallery/GalleryTile.tsx
kudlajz: I've tried fixing it with a timeout of 0, can you try it out? 7823d8411c2583e3251125722295849462ea78b3

packages/slate-editor/src/extensions/galleries/components/Gallery/SortableGalleryTile.tsx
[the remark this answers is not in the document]
kudlajz: Discussed this with @alpower and reworked the "drop" overlay to take the entire tile instead of appearing before/after. The drop behaviour is still the same, but it feels a bit more "natural" this way:

https://github.com/prezly/slate/assets/4209081/3d104066-86e5-4eca-ad0f-1ae0d1f6be0b

packages/slate-editor/src/components/EditorBlock/EditorBlock.tsx
e1himself: JFYI It's no longer possible to close the menu of the site gallery aside after this change.

https://github.com/user-attachments/assets/ad887935-4cd7-4bd6-8bf8-50a5036af5c0

e1himself: > That's because the gallery block is still selected. When deselected, the menu would close automatically.
However because there's no other node to select, you can't really deselect it.

Yep.

It's a little bit annoying :)

PR #529, opened 2024-02-26
[DEV-12566] Feature - Implement GalleryBookmark extension

kudlajz:
This is an extension similar to Story bookmarks that allows the user to select an existing gallery and creates a web bookmark for it with some predefined options (like layout).

### DIFF
--- a/packages/slate-editor/src/extensions/gallery-bookmark/GalleryBookmarkExtension.tsx
+++ b/packages/slate-editor/src/extensions/gallery-bookmark/GalleryBookmarkExtension.tsx
@@ -1,0 +1,7 @@
+import type { Extension } from '@prezly/slate-commons';
+
+export const EXTENSION_ID = 'GalleryBookmarkExtension';
+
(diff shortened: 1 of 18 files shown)


packages/slate-editor/src/extensions/gallery-bookmark/GalleryBookmarkExtension.tsx
```diff
@@ -0,0 +1,7 @@
+import type { Extension } from '@prezly/slate-commons';
+
+export const EXTENSION_ID = 'GalleryBookmarkExtension';
+
+export const GalleryBookmarkExtension = (): Extension => ({
```
kudlajz, line 5: The extension feels kinda useless here since it doesn't provide anything and we're using web bookmarks under the hood.

The only thing is the `createGalleryBookmark` function, but I was thinking if it would be easier to just place it inline to the `GalleryBookmarkPlaceholderElement` and get rid of this whole extension? 🤔

PR #534, opened 2024-03-11
[CARE-4588] Feature - Introduce more callbacks in gallery extension

kudlajz:
Introduce necessary callbacks in order to allow tracking various gallery node interactions.

### DIFF
--- a/packages/slate-editor/src/extensions/galleries/components/Gallery/Gallery.tsx
+++ b/packages/slate-editor/src/extensions/galleries/components/Gallery/Gallery.tsx
@@ -31,7 +31,11 @@ interface Props extends HTMLAttributes<HTMLDivElement> {
     images: GalleryImage[];
     isInteractive: boolean;
     maxViewportWidth?: number;
+    onImageCrop: (image: GalleryImage) => void;
+    onImageDelete: (image: GalleryImage) => void;
+    onImageEditCaption: (image: GalleryImage) => void;
     onImagesChange: (images: GalleryImage[]) => void;
+    onImagesReordered: (images: GalleryImage[]) => void;
     padding: GalleryNode['padding'];
     size: GalleryNode['thumbnail_size'];
     uuid: string;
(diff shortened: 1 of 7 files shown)


packages/slate-editor/src/extensions/galleries/components/Gallery/Gallery.tsx
```diff
@@ -162,6 +179,12 @@ export function Gallery({
     }
 
     function handleDelete(id: UniqueIdentifier) {
+        const imageToDelete = images.find((image) => image.id === id);
+        if (!imageToDelete) {
+            return;
+        }
+
+        onImageDelete(imageToDelete);
```
e1himself, line 187: My default approach is always to fire tracking/analytics callbacks after the business logic just to avoid analytics errors breaking the main logic.

Maybe we want it here too.

kudlajz: Yeah, I see what you mean. But a small problem here is that we're mainly interested in the click events, not when the operation finishes, e.g. caption is edited or image is cropped.

In theory, if I only fire `onImageCropped` at the end of the business logic, people that click "Crop" but then close the Uploadcare dialog will not fire this event, whereas what we want is fire the event for everyone as soon as they click "Crop".

That's why I went with present tense in some cases, otherwise I would just use past tense.

kudlajz: > How about onCropClicked? It's past tense, but clear that it's happening before the actual cropping?

Yeah, that works 👍

kudlajz: So, do you think it's okay if I just simplify some of the callbacks I've introduced into a respective `onClicked` callback? As we don't really care about the underlying properties or values, we simply care about the click events.

So, it would be:
- `onImageDeleteClicked`,
- `onImageCropClicked`
- `onImageCaptionClicked`

The remaining callbacks can stay in past tense as they are I think.

e1himself (approved):

packages/slate-editor/src/extensions/galleries/components/Gallery/Gallery.tsx
e1himself: Yeah, sounds good :+1:

PR #539, opened 2024-05-07
Fix build

e1himself:
So the tests started to fail ([even in master](https://github.com/prezly/slate/actions/runs/8987243389/job/24685238667)).

The error was caused by code using native `File` and `FormData` vs jsdom-mocked `window.File` and `window.FormData`, not compatible with the native implementations.

I've made the tests use the mocked `window.File` and `window.FormData` replacing globals. This fixes the problem.

I'm not sure though why it started to fail. Probably something has changed in one of the dependencies. As the master started to fail even without updating anything in the `package.json`


kudlajz (approved):
👏

MohammadxAli (approved):

PR #547, opened 2024-07-08
[CARE-5345] Fix - Always set the `text-align` property on paragraph

kudlajz:
Previously, if default alignment was applied to the paragraph (in our case `left`), it would not render the `text-align: left` style property.

This caused a bug in the tables, more specifically table header cell, which by default aligns the text to the center.

The solution is to explicitly render the `text-align` style property even if default alignment is used.

### DIFF
--- a/packages/slate-editor/src/extensions/paragraphs/components/ParagraphElement.tsx
+++ b/packages/slate-editor/src/extensions/paragraphs/components/ParagraphElement.tsx
@@ -1,4 +1,4 @@
-import { Alignment} from '@prezly/slate-types';
+import { Alignment } from '@prezly/slate-types';
 import { type ParagraphNode } from '@prezly/slate-types';
 import classNames from 'classnames';
 import type { HTMLAttributes } from 'react';
@@ -13,7 +13,14 @@ interface Props extends HTMLAttributes<HTMLParagraphElement> {
     element: ParagraphNode;
 }
 
-export function ParagraphElement({ attributes, children, className, defaultAlignment, element, ...props }: Props) {
+export function ParagraphElement({
+    attributes,
+    children,
+    className,
+    defaultAlignment,
+    element,
+    ...props
+}: Props) {
     const align = element.align || defaultAlignment;
 
     return (
--- a/packages/slate-editor/src/extensions/paragraphs/components/ParagraphElement.tsx
+++ b/packages/slate-editor/src/extensions/paragraphs/components/ParagraphElement.tsx
@@ -1,4 +1,5 @@
-import { Alignment, type ParagraphNode } from '@prezly/slate-types';
+import type { Alignment} from '@prezly/slate-types';
+import { type ParagraphNode } from '@prezly/slate-types';
 import classNames from 'classnames';
 import type { HTMLAttributes } from 'react';
 import React from 'react';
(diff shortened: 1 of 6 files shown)


packages/slate-editor/src/extensions/paragraphs/components/ParagraphElement.tsx
```diff
@@ -12,12 +12,14 @@ interface Props extends HTMLAttributes<HTMLParagraphElement> {
...
 export function ParagraphElement({ attributes, children, className, element, ...props }: Props) {
+    const align = element.align || Alignment.LEFT;
+
     return (
         <p
             {...attributes}
             {...props}
             className={classNames(styles.ParagraphElement, className)}
-            style={{ textAlign: element.align }}
+            style={{ textAlign: align }}
```
MohammadxAli, line 22: Isn't it better to use [CSS Logical Properties](https://caniuse.com/css-logical-props) for `text-align`? Using `text-align: start` is equal to `text-align: left` however, using `start` will ensure the RTL documents can benefit from the same fix.

MohammadxAli (approved):

packages/slate-editor/src/extensions/paragraphs/components/ParagraphElement.tsx
```diff
@@ -12,12 +12,14 @@ interface Props extends HTMLAttributes<HTMLParagraphElement> {
 }
 
 export function ParagraphElement({ attributes, children, className, element, ...props }: Props) {
+    const align = element.align || Alignment.LEFT;
```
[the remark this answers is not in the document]
e1himself: How would this work for RTL languages?

packages/slate-editor/src/extensions/paragraphs/components/ParagraphElement.tsx
e1himself: Hm... That's a good point. 
My main idea behind omitting `align` if it matches the default was to play well with i18n and RTL languages.

Changing the model, though, isn't easy.

packages/slate-editor/src/extensions/paragraphs/components/ParagraphElement.tsx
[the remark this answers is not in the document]
e1himself: Why not pass `defaultAlignment` to the `ParagraphsExtension`?

packages/slate-editor/src/extensions/paragraphs/components/ParagraphElement.tsx
kudlajz: So, based on Ali's point, I think it's better to use `text-align: start` and `text-align: end` for left and right alignment respectively as then it will work properly for RTL languages as well. However, the UI still shows the icons as left/center/right whereas it should probably be the other way around for RTL.

No need to change the model at this point, we just change the text align properties.

The [newsroom](https://ali-newsroom.prezly.com/translation-table) itself renders the alignment properly I believe.

packages/slate-editor/src/extensions/paragraphs/components/ParagraphElement.tsx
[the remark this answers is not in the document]
kudlajz: You're right, I've changed it and passing it now.

packages/slate-editor/src/extensions/paragraphs/components/ParagraphElement.tsx
kudlajz: Looks like this now, this is left, center and right align respectively:

LTR
Screenshot 2024-07-08 at 17 24 10

RTL
Screenshot 2024-07-08 at 17 23 36

So pretty much looks the same as in production right now, just with the bug resolved.

e1himself (approved):

PR #548, opened 2024-07-08
[CARE-5666] Fix - Use overflow: auto in tables

kudlajz:
Previously the table element was using `overflow: hidden`, now it will use `overflow: auto` showing a scrollbar if needed.

Screenshot 2024-07-08 at 17 46 54

### DIFF
--- a/packages/slate-editor/src/components/EditorBlock/EditorBlock.tsx
+++ b/packages/slate-editor/src/components/EditorBlock/EditorBlock.tsx
@@ -53,7 +53,7 @@ export interface Props
     layout?: `${Layout}`;
     loading?: boolean;
     menuPlacement?: PopperOptionsContextType['placement'];
-    overflow?: 'visible' | 'hidden';
+    overflow?: 'visible' | 'hidden' | 'auto';
     overlay?: OverlayMode;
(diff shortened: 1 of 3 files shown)


packages/slate-editor/src/components/EditorBlock/EditorBlock.tsx
```diff
@@ -203,6 +203,7 @@ export const EditorBlock = forwardRef<HTMLDivElement, Props>(function (
                         [styles.border]: border,
                         [styles.editable]: Boolean(renderEditableFrame),
                         [styles.hasError]: hasError,
+                        [styles.overflowAuto]: overflow === 'auto',
```
kudlajz, line 206: I was wondering if we need these classes or would it be simpler to just pass the `overflow` to the `style` property? 🤔 

We're using `style` in the `ParagraphElement` for alignment for example, so I felt like we could use it here as well since we're using the `overflow` property as-is.

MohammadxAli: Makes sense :)

MohammadxAli (approved):
👍

PR #550, opened 2024-07-09
[CARE-4591] Fix - Only allow URL bookmark autoformatting in empty paragraphs

kudlajz:


e1himself (approved):

packages/slate-editor/src/modules/editor/getEnabledExtensions.ts
```diff
@@ -448,12 +448,13 @@ function buildPlaceholdersExtensionConfiguration({
...
                               isAllowed(editor) {
-                                  const isSelectionEmpty = EditorCommands.isSelectionEmpty(editor);
+                                  const isSelectionInEmptyParagraph =
+                                      EditorCommands.isCursorInEmptyParagraph(editor);
                                   const isInsideTable = TablesEditor.isTablesEditor(editor)
                                       ? TablesEditor.isInTable(editor)
                                       : false;
 
-                                  return isSelectionEmpty && !isInsideTable;
+                                  return isSelectionInEmptyParagraph && !isInsideTable;
```
kudlajz, line 457: @e1himself I've removed the `isSelectionEmpty` call since `isCursorInEmptyParagraph` already makes sure range is not expanded (therefore collapsed).

I haven't noticed any difference in behaviour, so should be okay 🙂

e1himself: IDK, I'd personally keep it for clarity.

kudlajz: Yeah, makes sense, I've reverted it, thanks!

PR #554, opened 2024-07-16
[CARE-4611] Feature - Tweak various element menus

kudlajz:
- toolbox header now uses full opacity
- default (medium) button size changed to 14 px
  - introduced "small" and "tiny" sizes
  - removed `noPadding` variant
- "Remove X" changed to simply "Remove"
- various edit/view links shuffled to the bottom, above the "Remove" button

### DIFF
--- a/packages/slate-editor/src/components/Button/Button.module.scss
+++ b/packages/slate-editor/src/components/Button/Button.module.scss
@@ -10,16 +10,12 @@
     cursor: pointer;
     text-decoration: none;
     font-weight: 600;
-    font-size: $font-size-smaller;
+    font-size: $font-size-small;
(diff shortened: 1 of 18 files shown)


packages/slate-editor/src/components/Button/Button.module.scss
```diff
@@ -130,8 +126,23 @@
         border-radius: 6px;
     }
 
-    &--noPadding {
+    &--tiny {
         padding: 0;
+        font-size: $font-size-smaller;
+
+        .icon-wrapper {
```
kudlajz, line 133: Maybe this could be moved to `.icon-wrapper` as a modifier class instead.

PR #555, opened 2024-07-16
[CARE-5752] Feature - Implement settings menu for coverage cards

kudlajz:
Implements settings menu for coverage cards and tweaks its styling to somewhat mimic the bookmark cards.

Screenshot 2024-07-16 at 18 34 59


packages/slate-editor/src/extensions/coverage/lib/getCoverageImageUrl.ts
```diff
@@ -0,0 +1,22 @@
...
+const IMAGE_WIDTH = 580 * 2;
+
+export function getCoverageImageUrl(coverage: CoverageEntry): string | null {
+    if (coverage.attachment_oembed && coverage.attachment_oembed.thumbnail_url) {
+        return coverage.attachment_oembed.thumbnail_url;
+    }
+
+    if (coverage.attachment) {
+        try {
+            const image = UploadcareImage.createFromPrezlyStoragePayload(coverage.attachment as UploadedImage);
```
kudlajz, line 14: I've noticed the `attachment` is actually of type `UploadedFile` so the previous condition where we checked if it's an `UploadcareImage` failed, because the object doesn't have `effects` and other image properties, so thumbnails were not working at all.

I guess we still need that check because `attachment` can be a non-image file in which case it's not possible to display however I wonder why we're not using the `UploadedImage` structure for image files.

Is this a bug in the API? 🤔

MohammadxAli (approved):

fgyimah (approved):
👍🏾

packages/slate-editor/src/extensions/coverage/lib/getCoverageImageUrl.ts
e1himself: But it's `attachment`, not cover image. It can be a PDF file, MP3 or a video file.

kudlajz: Yeah, I've realised that, but the problem is that we were trying to render the `attachment` as a thumbnail if it's an image, but even if you upload an image itself, the API will return the payload without the `effects` and other image-specific properties so the condition will never be fulfilled.

e1himself: > but even if you upload an image itself, the API will return the payload without the effects and other image-specific properties so the condition will never be fulfilled.

Ah, that... In that case yeah, we can consider this as a bug in the API.

packages/slate-editor/src/extensions/coverage/components/CoverageElement.tsx
```diff
@@ -35,8 +37,9 @@ export function CoverageElement({
     dateFormat,
     element,
     fetchCoverage,
+    onEdit,
 }: Props) {
-    const editor = useSlateStatic();
+    const editor = useSlate();
```
e1himself, line 42: Just curious what was the reason for this change?

e1himself (approved):

packages/slate-editor/src/extensions/coverage/components/CoverageElement.tsx
kudlajz: I saw that some components were using `useSlate` and some were using `useSlateStatic` and I wasn't sure what's the difference there, so I made it consistent with the `CoverageMenu`.

However I've checked the docs and I see that `useSlateStatic` does not trigger a re-render when there are changes in the editor, as oposed to `useSlate` which does.

To be honest, I'm not really sure which one is the correct one to use? 🤔 I guess it's better to use `useSlateStatic` in most cases to prevent re-renders?

e1himself: If short, you want to use `useSlate()` if your component depends on the editor state (like `children` or `selection`) -- to re-render it when the state changes, and `useSlateStatic()` when all you want is imperative API (to query and modify editor state _in response to user actions_).

kudlajz: Ah, makes sense, thanks for explaining it 👍

packages/slate-editor/src/extensions/coverage/lib/getCoverageImageUrl.ts
kudlajz: Checked with @pestaa and turns out frontend was sending the `UploadcareFile` payload instead of `UploadcareImage` payload if attachment was an image 🥲 

So any past uploaded attachments will not have the image-specific properties until https://github.com/prezly/prezly/pull/15053 is deployed.

Either way, I think this should still work as the `UploadcareImage` component should throw an error if it's not possible to construct the object from the given payload.

I'll test this to make sure.

PR #558, opened 2024-07-18
[CARE-5694][CARE-1580] Feature - Implement improved version of [+] button

MohammadxAli:
## Mentions Dropdown (Removed `react-bootstrap` usage)

https://github.com/user-attachments/assets/d32bd5dd-d13c-4bd3-855f-389be25fd386

## New [+] menu

- Changed icons
- Improved styles
- Removed `react-bootstrap` usage

https://github.com/user-attachments/assets/0c522e0f-2003-4c41-a723-e3e64ac39572


babel.config.json
```diff
@@ -4,19 +4,42 @@
...
+                                    "overrides": {
+                                        "removeViewBox": false,
+                                        "cleanupIds": false
+                                    }
+                                }
+                            }
+                        ]
+                    }
+                }
+            }
```
MohammadxAli, line 26: This took me a while to understand how to properly deactivate cleaning up ids plugin of SVGO, if you're  interested to know why is this needed, please see: https://github.com/prezly/slate/pull/477#discussion_r1304336589

packages/slate-editor/.svgo.yml
```diff
@@ -1,2 +0,0 @@
-plugins:
```
MohammadxAli, line 1: This doesn't work, see my other comment.

packages/slate-editor/src/extensions/mentions/components/MentionsDropdown.tsx
```diff
@@ -97,3 +96,21 @@ export const MentionsDropdown = <V extends object>({
         </ul>
     );
 };
+
+function MenuItem({
+    active,
+    children,
+    className,
+    ...props
+}: ComponentPropsWithRef<'li'> & { active?: boolean }) {
```
e1himself, line 105: > `ComponentPropsWithRef`

Hm... I wonder if this is necessary. `ref` is not a normal property, and can be assigned to any element without declaring it.

Moreover, you will never receive the `ref` value here, as a property. So you want a way to forward the ref of the `li` it will not work this way. You either need `forwardRef()`, or use a different property name (like `innerRef`, or something).

e1himself (approved):

packages/slate-editor/src/extensions/mentions/components/MentionsDropdown.tsx
MohammadxAli: Ah good catch, there's also another type called [`ComponentPropsWithoutRef`](https://stackoverflow.com/a/73050903/6622233), I guess I'll need to use that.

e1himself: We usually use `HTMLAttributes` in such cases. Here I'd define it as `HTMLAttributes<HTMLLIElement>`.

Makefile
```diff
@@ -0,0 +1,14 @@
+.PHONY: watch clean fresh
+
+clean:
```
kudlajz, line 3: There's already a clean script in package.json, why do we have this one here and why is it different?

kudlajz (approved):

Makefile
MohammadxAli: It's not very different honestly, during linking/unlinking deleting `node_modules`, `.turbo` and `build` recursively is something I do a lot so I thought it's probably good to have it solved with a `make` task.

Basically it's the same but doesn't depend on `rimraf`. Should I remove it?

kudlajz: I don't know 🙂 I've just installed `rimraf` globally and then it's not a problem to run `pnpm clean` whenever needed. So not sure if we need a Makefile that kinda does the same thing 🤔

MohammadxAli: You're right :) Dropped it in favor of `npm` scripts.

PR #559, opened 2024-07-19
[CARE-5777] Fix - Change the [+] button to the "new" variants

MohammadxAli:
### Default

image

### Hover

image

### Open

image

---

### Default

image

### Hover

image

### Open

image


packages/slate-editor/src/extensions/floating-add-menu/components/Plus.module.scss
```diff
@@ -0,0 +1,18 @@
+@use "sass:math";
```
kudlajz, line 1: 🤔

kudlajz (approved):

PR #560, opened 2024-08-01
[CARE-5752] Feature - Make layout options optional in coverage settings

kudlajz:
I had to move the "Open in new tab" option to a separate section since it was previously part of the "Layout" section, which is now optional.

Screenshot 2024-08-01 at 13 47 32


fgyimah (approved):
👍🏾

PR #561, opened 2024-08-06
[DEV-13254] Implement settings menu for variables with optional fallback

kudlajz:
- variables no longer show as `%something%` but instead use the label for better clarity
- variables now have a menu when selected
- variables that have `withFallback: true` can have a text fallback set for them

I still have to run this with the UX people so the UI might change.

Default:
Screenshot 2024-08-06 at 19 00 53

With fallback defined:
Screenshot 2024-08-06 at 19 01 03

Choosing different options:
Screenshot 2024-08-06 at 18 36 52

### DIFF
--- a/packages/slate-editor/src/extensions/variables/VariableElement.module.scss
+++ b/packages/slate-editor/src/extensions/variables/VariableElement.module.scss
@@ -8,8 +8,8 @@
      *
      * Also, it's not possible to drag a node when it's selectable.
      *
-     * see: https://app.clubhouse.io/prezly/story/19824/weird-selection-after-editing-an-attachment-or-image-or-gallery
-     * see: https://app.clubhouse.io/prezly/story/20456/error-cannot-resolve-a-dom-node-from-slate-node-text
+     * see: https://github.com/prezly/clubhouse-archive/blob/master/archive/stories/19824.md
+     * see: https://github.com/prezly/clubhouse-archive/blob/master/archive/stories/20456.md
      */
     user-select: none;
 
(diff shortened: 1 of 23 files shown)


packages/slate-editor/src/extensions/variables/VariableElement.module.scss
```diff
@@ -0,0 +1,29 @@
...
+
+.VariableElement {
+    /*
+     * Selection in Slate 0.50+ does not work as expected if a void block is selectable.
+     * It could lead to an error when Slate is resolving the node which crashes the editor.
+     *
+     * Also, it's not possible to drag a node when it's selectable.
+     *
+     * see: https://app.clubhouse.io/prezly/story/19824/weird-selection-after-editing-an-attachment-or-image-or-gallery
+     * see: https://app.clubhouse.io/prezly/story/20456/error-cannot-resolve-a-dom-node-from-slate-node-text
```
e1himself, line 12:
```suggestion
     * see: https://github.com/prezly/clubhouse-archive/blob/master/archive/stories/19824.md
     * see: https://github.com/prezly/clubhouse-archive/blob/master/archive/stories/20456.md
```

e1himself (approved):

packages/slate-types/src/nodes/VariableNode.ts
```diff
@@ -5,6 +5,7 @@ export const VARIABLE_NODE_TYPE = 'variable';
 
 export interface VariableNode extends ElementNode {
     type: typeof VARIABLE_NODE_TYPE;
+    fallback?: string | null;
```
e1himself, line 8: @kudlajz by the way, Slate does not allow `null` as a node property value. Setting a property to null with `Transforms.setNodes()` will remove it from the object.

PR #562, opened 2024-08-08
[DEV-13267] Make fallback prop not nullable, remove it if it's empty

kudlajz:


packages/slate-editor/src/extensions/variables/VariableMenu.tsx
```diff
@@ -48,10 +48,6 @@ export function VariableMenu({ container, element, onClose, variables }: Props)
 
     useEffect(() => {
         if (!option?.withFallback) {
-            // Even though the API supports `null` for the fallback,
-            // we can't use it here because Slate will remove this property
-            // and the API will add it back upon saving, creating an endless loop
-            // since the two values will differ.
             updateVariable(editor, { fallback: '' });
         }
     }, [option?.withFallback]);
```
e1himself, line 53: Can we not just remove this effect? Why is it necessary? 
Can it be replaced with a normalization function in the VariablesExtension?

e1himself (approved):

packages/slate-editor/src/extensions/variables/VariableMenu.tsx
kudlajz: I believe it is needed because you can change the variable type in the menu and if you change to a type that does not allow a fallback, the previous fallback value (if defined) would stay in the node.

This way, the fallback value will get removed.

e1himself: I believe we sometimes abuse effects for what they are not designed for.

https://react.dev/learn/synchronizing-with-effects#you-might-not-need-an-effect

PR #565, opened 2024-08-08
[CARE-5911] Introduce "type" dropdown into the link menu, allow predefined values as links

kudlajz:
Allow choosing what kind of link the user is trying to add (web is the default):
Screenshot 2024-08-08 at 17 27 02

If email is choosed, the email address will be normalized upon saving into `mailto:your@email.com`.
When editing an email link, it is "humanized" before being presented in the input so the `mailto:` gets stripped.
Screenshot 2024-08-08 at 17 27 19
Screenshot 2024-08-08 at 17 27 23

Also allow selecting predefined values (variables) as link hrefs, used for the new unsubscribe URL and also for the already implemented story and short story URLs:
Screenshot 2024-08-08 at 17 27 32

The link tooltip reflects this and shows the human variable name (previously it would show a clickable `%release.url%` leading nowhere):
Screenshot 2024-08-08 at 17 27 38

### DIFF
--- a/packages/slate-editor/src/extensions/inline-links/types.ts
+++ b/packages/slate-editor/src/extensions/inline-links/types.ts
@@ -1,0 +1,6 @@
+export interface InlineLinksExtensionConfiguration {
+    predefinedLinks?: {
+        label: string;
+        options: { label: string; value: string }[];
+    };
+}
--- a/packages/slate-editor/src/extensions/inline-links/InlineLinksExtension.tsx
+++ b/packages/slate-editor/src/extensions/inline-links/InlineLinksExtension.tsx
@@ -23,7 +23,7 @@ export const EXTENSION_ID = 'InlineLinksExtension';
 
 export const InlineLinksExtension = ({
     predefinedLinks,
-}: InlineLinksExtensionConfiguration): Extension => ({
+}: InlineLinksExtensionConfiguration = {}): Extension => ({
     id: EXTENSION_ID,
     deserialize: {
         element: composeElementDeserializer({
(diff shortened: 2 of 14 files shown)


packages/slate-editor/src/extensions/inline-links/types.ts
```diff
@@ -0,0 +1,6 @@
+export interface InlineLinksExtensionConfiguration {
+    predefinedLinks?: {
```
kudlajz, line 2: I was thinking about making this into an array of objects but given our current requirements there's no need.

kudlajz: Maybe we should call it `predefinedHrefs`? 🤔 Not sure.

e1himself (approved):

packages/slate-editor/src/extensions/inline-links/types.ts
e1himself: What is the value that will be passed to this prop from the app?

kudlajz:
```
predefinedLinks: {
    label: 'Prezly content',
    options: [
        { label: 'Link to unsubscribe', value: '%unsubscribe_url%' },
        { label: 'Link to story', value: '%release.url%' },
        { label: 'Short link to story', value: '%release.shorturl%' },
    ],
}
```

e1himself: Looks fine as is, if you ask me :)

PR #569, opened 2024-09-17
[CARE-6013] Feature - Refetch coverage when update event happens

kudlajz:


packages/slate-editor/src/extensions/coverage/components/CoverageElement.tsx
```diff
@@ -60,6 +60,16 @@ export function CoverageElement({
...
 
+    function handleCoverageUpdated({ coverage }: { coverage: Pick<CoverageEntry, 'id'>}) {
+        if (element.coverage.id === coverage.id) {
+            loadCoverage();
+        }
+    }
+
+    useEffect(() => {
+        EventsEditor.addEventListener(editor, 'coverage-updated', handleCoverageUpdated);
+    }, []);
```
e1himself, line 71: 1. Shouldn't `handleCoverageUpdated` be a local function to the effect callback?
2. Should it not remove the event listener in the effect unmount callback?

e1himself (approved):

packages/slate-editor/src/extensions/coverage/components/CoverageElement.tsx
kudlajz: You're right, I forgot 🙂

PR #573, opened 2024-11-01
[DEV-13410] Fix - Coverage placeholder search input

kudlajz:
I've noticed some problems in the app with the usage of Popper, so I've brought back the minHeight and maxHeight calculation, so Popper doesn't have to flip the entire dropdown unless the available space is less than minHeight.

This makes the dropdown still usable while opened near the bottom of the screen.

There was also a bug with the position of the dropdown when addon is present - this was due to usage of `bottom` placement which anchors to the center of the element (the search input). I've changed it to anchor to `bottom-end`, so it's always aligned.

Another thing that has been fixed is the active element is now properly visible when the dropdown is displayed (previously the active item was sometimes hidden, since the dropdown was scrolled down a bit).

Screenshot 2024-11-01 at 15 53 04
Screenshot 2024-11-01 at 15 53 10
Screenshot 2024-11-01 at 15 53 33

### DIFF
--- a/packages/slate-editor/src/components/SearchInput/Suggestions.tsx
+++ b/packages/slate-editor/src/components/SearchInput/Suggestions.tsx
@@ -53,7 +53,7 @@ export function Suggestions<T>({
         placement: 'bottom-end',
     });
 
-    const updatePanelSizeAndPosition = useFunction(() => {
+    const updatePanelSize = useFunction(() => {
         setHeight(childrenContainer.current?.getBoundingClientRect().height);
 
         if (container.current) {
@@ -65,18 +65,22 @@ export function Suggestions<T>({
         }
     });
 
-    useEffect(updatePanelSizeAndPosition, [
-        query,
-        suggestions,
-        calculatedMaxHeight,
-        minHeight,
-        maxHeight,
-    ]);
+    useEffect(() => {
+        window.addEventListener('scroll', updatePanelSize);
+        window.addEventListener('resize', updatePanelSize);
+
+        return () => {
+            window.removeEventListener('scroll', updatePanelSize);
+            window.removeEventListener('resize', updatePanelSize);
+        };
+    }, [updatePanelSize]);
+
+    useEffect(updatePanelSize, [query, suggestions, minHeight, maxHeight]);
 
     useEffect(() => {
         async function repositionPopper() {
             await popper.update?.();
-            updatePanelSizeAndPosition();
+            updatePanelSize();
 
             if (activeElement) {
                 scrollarea?.ensureVisible(activeElement);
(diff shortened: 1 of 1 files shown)


mohammadxali (approved):
👍

fgyimah (approved):

packages/slate-editor/src/components/SearchInput/Suggestions.tsx
```diff
@@ -44,42 +46,51 @@ export function Suggestions<T>({
...
+        placement: 'bottom-end',
     });
 
     const updatePanelSizeAndPosition = useFunction(() => {
         setHeight(childrenContainer.current?.getBoundingClientRect().height);
+
+        if (container.current) {
+            const viewport = document.body.getBoundingClientRect();
+            const rect = container.current.getBoundingClientRect();
+            setMaxHeight(clamp(viewport.height - rect.top - 4, minHeight, maxHeight));
```
yuriyyakym, line 62: What is `4` magic number? 🤔

packages/slate-editor/src/components/SearchInput/Suggestions.tsx
```diff
@@ -44,42 +46,51 @@ export function Suggestions<T>({
...
+        } else {
+            setMaxHeight(undefined);
+        }
     });
 
-    useEffect(updatePanelSizeAndPosition, [query, suggestions, maxHeight]);
+    useEffect(updatePanelSizeAndPosition, [
```
yuriyyakym, line 68: Good improvement, but this should also happen on window resize and other events that may change available space.
Not sure if it's needed to be addressed in this PR.

yuriyyakym (approved):

packages/slate-editor/src/components/SearchInput/Suggestions.tsx
kudlajz: I'm not sure but I think it's probably a bottom padding. I restored this code from how it was before.
I can add a constant for it, but I'm gonna rework these dialogues this week during [my project](https://linear.app/prezly/project/refactor-placeholders-in-slate-editor-237d6ef14225) so probably not worth spending extra time on this.

packages/slate-editor/src/components/SearchInput/Suggestions.tsx
```diff
@@ -44,42 +46,51 @@ export function Suggestions<T>({
...
+        query,
+        suggestions,
+        calculatedMaxHeight,
```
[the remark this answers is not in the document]
kudlajz: Good catch, I think I just forgot to remove it after introducing the `repositionPopper` effect. I'll remove it.

PR #574, opened 2024-11-05
[DEV-13852] Refactoring - Allow custom rendering of placeholders

kudlajz:
The main idea behind this refactoring is to simplify the rendering of placeholders, allowing full control over the contents of the placeholders as soon as they're clicked and to also improve the developer experience when adding or modifying the existing UI. Since a lot of the placeholder code will now be part of the consumer codebase, we don't have to release new versions of the editor when doing copy changes for example.

I chose the `GalleryBookmarkPlaceholderElement` as a proof of concept since it also contains the site switcher. The plan is (if we agree to it) to rework the remaining "search input" placeholders and then eventually rework the regular "input" placeholders as well.

The interface has been simplified into sort of a "headless" approach, so now the consumer only needs to provide a `renderPlaceholder` function that will receive different props based on the required logic.

This allows full control over the UI and also allows usage of consumer code's UI components instead of having to rely on the UI components present (or not present) in the editor. This way, we can get rid of things like `invalidateSuggestions` etc. that were more of a band-aid to the current solution rather than a proper fix.

In this case, the gallery bookmark placeholder rendering function will receive props like `onSelect` (when an option is selected) and `onRemove` (when we want to remove the placeholder).

For example, the `onSelect` callback expects a promise that will resolve to an object containing the `OEmbedInfo` of a given gallery, so we no longer need to provide `fetchOembed` to the editor and we can do all of this outside in the app.

You can see how the new approach is implemented in https://github.com/prezly/prezly/pull/15756.

Please also read the comment below and provide your thoughts on this approach.

### DIFF
--- a/packages/slate-editor/src/extensions/placeholders/elements/GalleryBookmarkPlaceholderElement.tsx
+++ b/packages/slate-editor/src/extensions/placeholders/elements/GalleryBookmarkPlaceholderElement.tsx
@@ -83,7 +83,11 @@ export function GalleryBookmarkPlaceholderElement({
             overflow="visible"
             renderFrame={
                 isCustomRendered
-                    ? () => renderPlaceholder({ onRemove: handleRemove, onSelect: handleSelect })
+                    ? () =>
+                          renderPlaceholder({
+                              onRemove: removable ? handleRemove : undefined,
+                              onSelect: handleSelect,
+                          })
                     : undefined
             }
             removable={removable}
@@ -98,7 +102,7 @@ export namespace GalleryBookmarkPlaceholderElement {
         extends Pick<BaseProps<NewsroomGallery>, 'attributes' | 'children' | 'element' | 'format'>,
             Pick<PlaceholderElementProps, 'removable'> {
         renderPlaceholder: (props: {
-            onRemove: () => void;
+            onRemove: (() => void) | undefined;
             onSelect: (promise: Promise<{ oembed?: OEmbedInfo; url: string }>) => void;
         }) => ReactNode;
     }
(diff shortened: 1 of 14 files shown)


packages/slate-editor/src/extensions/placeholders/elements/GalleryBookmarkPlaceholderElement.tsx
```diff
@@ -72,61 +66,45 @@ export function GalleryBookmarkPlaceholderElement({
...
-            inputDescription="Add a media gallery card to your stories, campaigns and pitches"
-            inputPlaceholder="Search for media galleries"
-            onSelect={handleSelect}
+            onClick={() => setCustomRendered(true)}
+            overflow="visible"
+            renderFrame={
+                isCustomRendered
+                    ? () => renderPlaceholder({ onRemove: handleRemove, onSelect: handleSelect })
+                    : undefined
+            }
```
kudlajz, line 93: I am still wondering if it's a good idea to have this half-approach where the "initial" placeholder UI is rendered by the editor, but then the subsequent UI is custom rendered.

I think ideally it would be nice if we could render the whole placeholder outside of the editor (i.e. rendered via `renderPlaceholder` function) but this would also mean that we would need to implement drag & drop functionality and other things (loading states, placeholder management etc.), so I wasn't sure if we should do it now or at all.

My thinking was that the placeholder extension would be a "headless" extension, only providing the necessary logic (callbacks etc.), but then expecting the consumer to implement the UI for it.

What do you think?

e1himself (approved):

fgyimah (approved):
👍🏾

PR #580, opened 2025-01-15
[DEV-19139] Refactor - Drop `Extension.isElementEqual`

e1himself:


packages/slate-editor/src/modules/editor/plugins/withElementsEqualityCheck.ts
```diff
@@ -15,20 +29,100 @@ export interface ElementsEqualityCheckEditor {
...
-                    return ret;
-                }
-            }
-            return undefined;
+export function withElementsEqualityCheck<T extends SlateEditor>(
+    editor: T,
+): T & ElementsEqualityCheckEditor {
+    return Object.assign(editor, {
+        isElementEqual,
+    });
```
e1himself, line 37: Potentially the function could have just become a lib function (and not a method of `editor`). But I wanted to make the minimal necessary changes at this stage.

kudlajz (approved):

